PR #163, opened 2025-08-21
liquidity-provider: add unstake functionality

qmsiclghzzz666:


programs/liquidity-provider/src/lib.rs
```diff
@@ -443,6 +661,80 @@ pub struct ClaimGt<'info> {
...
+
+    /// Vault holding staked LP tokens (PDA)
+    #[account(
+        mut,
+        seeds = [
+            POSITION_SEED,
+            global_state.key().as_ref(),
+            owner.key().as_ref(),
+            &position_id.to_le_bytes(),
+            b"vault",
```
orion-047, line 705: Same as the `StakeLp` case.

PR #165, opened 2025-08-25
liquidity-provider: add apy_gradient functionality

qmsiclghzzz666:


programs/liquidity-provider/src/lib.rs
```diff
@@ -35,18 +33,21 @@ pub mod gmsol_liquidity_provider {
     use super::*;
 
     /// Initialize LP staking program
-    pub fn initialize(ctx: Context<Initialize>, min_stake_value: u128) -> Result<()> {
+    pub fn initialize(
+        ctx: Context<Initialize>,
+        min_stake_value: u128,
+        apy_gradient: [u128; 53], // each item is 1e20-scaled APR for week buckets [0-1), [1-2), ..., [52, +inf)
```
orion-047, line 39: Passing the full list directly may exceed the transaction size limit (the parameter size is 848 bytes), especially in scenarios executed through a Squads Vault Transaction. It's recommended to consider a partial update approach or passing a sparse table instead.

programs/liquidity-provider/src/lib.rs
```diff
@@ -394,6 +367,17 @@ pub mod gmsol_liquidity_provider {
         Ok(())
     }
 
+    /// Update APY gradient array (53 buckets)
+    pub fn update_apy_gradient(
+        ctx: Context<UpdateApyGradient>,
+        new_apy_gradient: [u128; 53],
```
orion-047, line 373: Same as the `initialize` instruction.

programs/liquidity-provider/src/lib.rs
```diff
@@ -507,6 +501,43 @@ fn compute_reward_with_cpi<'info>(
...
+            apy_gradient[week_index]
+        } else {
+            apy_gradient[52]
+        };
+        // accumulate apy * seconds
+        acc = acc.saturating_add(apy.saturating_mul(segment_seconds));
+        remaining = remaining.saturating_sub(segment_seconds);
+        current_start = week_end;
+        week_index = week_index.saturating_add(1);
+    }
```
orion-047, line 531: This piece of code looks like a modulo division.

orion-047 (changes requested):

orion-047 (approved):

PR #222, opened 2025-09-18
cli: Add support for querying updatable market config

thomas-pixel-star:


crates/cli/src/commands/market.rs
```diff
@@ -315,6 +315,20 @@ enum Command {
         #[arg(required = true, num_args = 1..)]
         addresses: Vec<Pubkey>,
     },
+    /// Display updatable market config flags and/or factors.
+    UpdatableConfig {
+        #[arg(long)]
+        flags: bool,
+        #[arg(long)]
+        factors: bool,
```
orion-047, line 323: Consider changing to `without_flags` and `without_factors` (i.e., show all by default).

crates/cli/src/commands/market.rs
```diff
@@ -315,6 +315,20 @@ enum Command {
...
+    },
+    /// Query if a specific market config flag or factor is updatable.
+    IsConfigUpdatable {
+        #[arg(long, group = "what")]
+        flag: Option<String>,
```
orion-047, line 328: Use `MarketConfigFlag` as the type

crates/cli/src/commands/market.rs
```diff
@@ -315,6 +315,20 @@ enum Command {
...
+        #[arg(long, group = "what")]
+        factor: Option<String>,
```
orion-047, line 330: Use `MarketConfigKey` as the type

crates/cli/src/commands/market.rs
```diff
@@ -820,6 +834,105 @@ impl super::Command for Market {
...
+                };
+                use gmsol_sdk::programs::gmsol_store::accounts as store_accounts;
+                use std::convert::TryFrom;
+
+                let store_account: std::sync::Arc<store_accounts::Store> =
+                    client.store(store).await?;
+
+                if *flags || !(*factors) {
+                    let mut rows = Vec::new();
+                    for i in u8::MIN..=u8::MAX {
```
orion-047, line 849: Prefer using `MarketConfigFlag::iter()` directly for iteration

crates/cli/src/commands/market.rs
```diff
@@ -820,6 +834,105 @@ impl super::Command for Market {
...
+                            ("name", "Flag"),
+                            ("updatable", "Updatable"),
+                        ]),
+                    )?;
+                    println!("{out}");
+                }
+
+                if *factors || !(*flags) {
+                    let mut rows = Vec::new();
+                    for i in u16::MIN..=u16::MAX {
```
orion-047, line 874: Prefer using `MarketConfigKey::iter()` directly for iteration.

orion-047 (changes requested):
You can replace
```rust
(factors.value & (1u128 << idx)) != 0
```
with `get_flag` after #223 is merged.

crates/cli/src/commands/market.rs
thomas-pixel-star: changed

crates/cli/src/commands/market.rs
thomas-pixel-star: changed

crates/cli/src/commands/market.rs
thomas-pixel-star: changed

crates/cli/src/commands/market.rs
thomas-pixel-star: changed

orion-047 (approved):

PR #265, opened 2025-10-28
Update .gitignore to exclude .so files

mubarizkyc:


orion-047:
Thanks for the contribution!

However, these `.so` files are needed for testing (test-validator), so we shouldn't remove them.

mubarizkyc:
check [this](https://github.com/LiteSVM/litesvm) as well

mubarizkyc:
litesvm v7 works with solana v2.2.1

orion-047:
Thank you for sharing the example! From our side, having a separate test project within the same repo should work fine—as long as it isn't included in the current workspace, we shouldn't run into dependency conflicts.

Also, it looks like we may not need to manually define those interfaces in `store_utils`. Since we can use `anchor_lang::declare_program!`, the interfaces can be generated automatically from the IDL.

PR #302, opened 2026-01-12
Create rust.yml

francojaf:
name: Cache cargo build
  uses: actions/cache@v4
  with:
    path: |
      ~/.cargo/registry
      ~/.cargo/git
      target
    key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
    restore-keys: |
      ${{ runner.os }}-cargo-


angelo-75:
The purpose of this PR is unclear, so closing it.

PR #386, opened 2026-07-20
test: GLV deposit market-count ramp — execution fails at 26 markets (CU limit)

plutohan:
## What

Adds a **many-markets GLV** (index 1) to the anchor test deployment with 40 candidate markets backed by synthetic index tokens (reusing the SOL Pyth feed, so the oracle side stays flat), plus a new test `glv_deposit_with_increasing_markets` that inserts markets into the GLV **one by one**, executing a GLV deposit at every count to find where execution breaks.

The primary 2-market GLV and the existing `glv_deposit` test are unchanged; the ramp uses its own GLV.

## Findings (localnet, pinned toolchain)

- GLV deposit execution succeeds at **every count from 2 to 25 markets**.
- It **fails at 26 markets** by exhausting the compute budget: `ExecuteGlvDeposit` consumes **949,265 CU of the 999,700 granted** at 25 markets, and each additional market costs ~35–40k CU in GLV valuation. At 26 the instruction gets only the transaction's remaining ~50k CU and dies mid-execution.
- Transaction size and account locks are not the binding constraint (the execute path uses v0 transactions with the market ALT).
- Extrapolation: raising the compute budget toward the 1.4M cap would buy roughly 10 more markets (~35); beyond that a single-transaction execution cannot value the GLV, well below the state-level cap of 96 markets.

The test does not fail when it finds the limit — it logs the last successful count and the error (visible with `GMSOL_TEST_LOG=error,anchor=info` and `EXTRA_CARGO_ARGS=--nocapture`).

## Supporting changes

- Test users are funded with 20 SOL (the keeper pays rent for the 40 candidate markets, which no longer fits in 2 SOL).
- The market ALT also covers the many-markets GLV addresses (GLV, GLV token, per-market vault ATAs).
- `RUST_LOG` is scoped to the test process inside the Anchor.toml test script: exporting it in the environment reaches the spawned `solana-test-validator` and breaks its startup (set `GMSOL_TEST_LOG` instead).

## Verification

- `cargo check` clean; GLV test group green (6/6) including the reverted original `glv_deposit`.
- Full suite run in progress at time of writing; the 24-market variant of this setup was fully green (39 passed + the 2 pre-existing flaky tests passing on isolated retry).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Anchor.toml
+++ b/Anchor.toml
@@ -25,7 +25,10 @@ cluster = "localnet"
 wallet = "~/.config/solana/id.json"
 
 [scripts]
-test = "cargo test --test anchor --features anchor-test -- $GMSOL_TEST $EXTRA_CARGO_ARGS"
+# `RUST_LOG` is scoped to the test process here on purpose: exporting it in the
+# environment would also reach the spawned `solana-test-validator` and break its
+# startup. Set `GMSOL_TEST_LOG` (e.g. to `error,anchor=info`) to see the test logs.
+test = "RUST_LOG=${GMSOL_TEST_LOG:-error} cargo test --test anchor --features anchor-test -- $GMSOL_TEST $EXTRA_CARGO_ARGS"
 
 [test]
 startup_wait = 15000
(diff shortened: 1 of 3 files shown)


plutohan:
Closing — this was exploratory measurement work for an analysis report, not intended for merge.

PR #390, opened 2026-07-27
feat: expose market-status flags in SDK/CLI

gmitsj:
v0.10.0 added a per-feed market-status policy (`MarketStatusFlag` / `FeedConfig::market_status_flags`, a `u8` bitmap resolving how each `MarketStatus` reading resolves to open/closed) with a MARKET_KEEPER-gated setter, but no read-side visibility: the raw bitmap had no serde support, and neither the SDK's token-map view nor the CLI surfaced it. Users could set the policy but not see it. This PR closes that gap.

## What this adds

- **`MarketStatusFlag` is now serializable and iterable** (`crates/utils/src/price/market_status.rs`): `serde` (snake_case, matching the existing `clap::ValueEnum` naming so CLI input and JSON output agree), `strum::Display` (snake_case), and `strum::EnumIter` behind the existing `enum-iter` feature — the same gating pattern already used for `PriceProviderKind`.
- **`SerdeFeedConfig` gains `market_status_flags: u8`** (`crates/sdk/src/serde/serde_token_map.rs`), populated from `FeedConfig::market_status_flags().into_value()` in both match arms of `from_feed_config`. `#[serde(default)]` so existing serialized token-map dumps without the field still deserialize.
- **CLI decodes the bitmap for display** (`crates/cli/src/commands/market.rs`): `TokenConfigDisplay` wraps `SerdeTokenConfig` (via `#[serde(flatten)]`) and adds a per-provider `market_status: IndexMap<PriceProviderKind, String>` of decoded flag names, e.g. `"allow_pre_market, halt_regular_hours"`, rendered as `"-"` when no flags are set. Wired into the existing `market_status.<provider>` keys in the account-display output alongside the feed/timestamp-adjustment columns.

## Design decision: raw bitmap in the SDK, decoding only in the CLI (review focus)

`SerdeFeedConfig` deliberately exposes the **raw `u8`**, not a decoded flag list. `MarketStatusFlag` is `#[non_exhaustive]` and append-only over a live bitmap (each variant's position is a persisted bit), so a future flag added on-chain would be silently dropped by any SDK-level decode step running an older build — exactly the kind of round-trip data loss the SDK's serde views elsewhere avoid (see `max_deviation_factor`'s own `#[serde(default)]` precedent). Keeping the SDK layer as the raw, lossless value pushes decoding to the CLI's display-only `TokenConfigDisplay`, which explicitly reports bits it doesn't recognize as `"unknown (bit N)"` rather than dropping them — so an out-of-date CLI degrades legibly instead of hiding information.

## Tests

- `market_status.rs`: `Display`/serde snake_case naming per variant; JSON round-trip; `EnumIter` yields all variants in declaration order.
- `serde_token_map.rs`: bitmap round-trips through JSON for zero, a single flag, combined flags, and an unknown high bit (constructed directly via the container, bypassing the enum, to simulate a future flag this build doesn't recognize); `from_feed_config` surfaces the flags identically in both provider match arms, and defaults to zero when unset.
- `commands/market.rs`: `decode_market_status_flags` unit tests (zero, single flag, combined flags, unknown-bit-only, known+unknown mixed) and one end-to-end test that `TokenConfigDisplay` decodes per-provider flags from a constructed `SerdeTokenConfig`.

## Compatibility

Additive and backward compatible: new field is `#[serde(default)]` on the SDK struct, and the CLI change only adds output columns. No account layout or instruction changes.

### DIFF
--- a/crates/cli/src/commands/market.rs
+++ b/crates/cli/src/commands/market.rs
@@ -28,7 +28,7 @@ use gmsol_sdk::{
     },
     serde::{
         serde_market::{SerdeMarket, SerdeMarketConfig, SerdeMarketConfigBuffer},
-        serde_token_map::SerdeTokenConfig,
+        serde_token_map::{SerdeMarketStatusFlags, SerdeTokenConfig},
         StringPubkey,
     },
     solana_utils::{
@@ -424,7 +424,7 @@ impl super::Command for Market {
 
                 if let Some(token) = token {
                     let config = token_map.get(token).ok_or_eyre("token not found")?;
-                    let serialized = SerdeTokenConfig::try_from(config)?;
+                    let serialized = TokenConfigDisplay::new(SerdeTokenConfig::try_from(config)?);
                     println!(
                         "{}",
                         output.display_keyed_account(
@@ -443,13 +443,16 @@ impl super::Command for Market {
                                     "feeds.chainlink_data_streams.timestamp_adjustment",
                                     "Chainlink TS Adj",
                                 ),
+                                ("market_status.chainlink_data_streams", "Chainlink Status",),
                                 ("feeds.pyth.feed_id", "Pyth Feed"),
                                 ("feeds.pyth.timestamp_adjustment", "Pyth TS Adj",),
+                                ("market_status.pyth", "Pyth Status"),
                                 ("feeds.switchboard.feed_id", "Switchboard Feed"),
                                 (
                                     "feeds.switchboard.timestamp_adjustment",
                                     "Switchboard TS Adj",
                                 ),
+                                ("market_status.switchboard", "Switchboard Status",),
                             ])
                         )?
                     );
@@ -1690,6 +1693,59 @@ struct TokenMetadata {
     uri: String,
     #[serde(default)]
     init: bool,
+}
+
+/// [`SerdeTokenConfig`] augmented with a joined, human-readable summary of
+/// each feed's market-status flags, for CLI display only.
+#[derive(serde::Serialize)]
+struct TokenConfigDisplay {
+    #[serde(flatten)]
+    config: SerdeTokenConfig,
+    market_status: IndexMap<PriceProviderKind, String>,
+}
+
+impl TokenConfigDisplay {
+    fn new(config: SerdeTokenConfig) -> Self {
+        let market_status = config
+            .feeds
+            .iter()
+            .map(|(kind, feed)| (*kind, market_status_summary(&feed.market_status)))
+            .collect();
+        Self {
+            config,
+            market_status,
+        }
+    }
+}
+
+/// Join the set `allow_*` flags into a comma-separated summary, e.g.
+/// `"allow_unknown, halt_regular_hours"`, or `"-"` if none are set.
+fn market_status_summary(flags: &SerdeMarketStatusFlags) -> String {
+    let mut names = Vec::new();
+    if flags.allow_unknown {
+        names.push("allow_unknown");
+    }
+    if flags.allow_pre_market {
+        names.push("allow_pre_market");
+    }
+    if flags.halt_regular_hours {
+        names.push("halt_regular_hours");
+    }
+    if flags.allow_post_market {
+        names.push("allow_post_market");
+    }
+    if flags.allow_overnight {
+        names.push("allow_overnight");
+    }
+    if flags.allow_closed {
+        names.push("allow_closed");
+    }
+
+    if names.is_empty() {
+        "-".to_string()
+    } else {
+        names.join(", ")
+    }
 }
 
 #[derive(Debug, serde::Serialize)]
--- a/crates/cli/src/commands/market.rs
+++ b/crates/cli/src/commands/market.rs
@@ -443,16 +443,19 @@ impl super::Command for Market {
                                     "feeds.chainlink_data_streams.timestamp_adjustment",
                                     "Chainlink TS Adj",
                                 ),
-                                ("market_status.chainlink_data_streams", "Chainlink Status",),
+                                (
+                                    "market_status_summary.chainlink_data_streams",
+                                    "Chainlink Status",
+                                ),
                                 ("feeds.pyth.feed_id", "Pyth Feed"),
                                 ("feeds.pyth.timestamp_adjustment", "Pyth TS Adj",),
-                                ("market_status.pyth", "Pyth Status"),
+                                ("market_status_summary.pyth", "Pyth Status"),
                                 ("feeds.switchboard.feed_id", "Switchboard Feed"),
                                 (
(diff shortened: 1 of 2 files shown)


gmitsj:
The ticket's description names both "detailed/list views" as missing the market-status policy today. This PR fully covers the detail view (market tokens <token>), decoded flag names per provider, plus the raw bitmap for free in JSON/TOML output. It does not add decoded flags to the list view's table (market tokens, no token specified).

Reasoning: the list view's table already has 7 columns (Name, Pubkey, Enabled, Synthetic, Decimals, Price Precision, Expected Provider). Adding one column per provider (Pyth/Chainlink/Switchboard), mirroring what the detail view does, pushes it to 10 columns and roughly doubles the table's width, and in practice most tokens only configure one provider, so two of the three new columns render as "-" for nearly every row. I rendered both versions side by side to check before deciding, and the 10-column version was materially harder to scan for what's normally a quick "list all tokens" glance.

crates/utils/src/price/market_status.rs
```diff
@@ -41,12 +41,16 @@ pub enum MarketStatus {
 /// the only hard floor — a stale price is always closed regardless of any flag.
 /// (`AllowClosed` on a v8 feed is generally not advised, but staleness still
 /// backstops it.)
-#[derive(Clone, Copy, PartialEq, Eq, IntoPrimitive, TryFromPrimitive)]
+#[derive(strum::Display, Clone, Copy, PartialEq, Eq, IntoPrimitive, TryFromPrimitive)]
+#[strum(serialize_all = "snake_case")]
```
angelo-75: Please explain the necessity of introducing `strum` here.

Generally, we try to avoid changes that affect the contract layer due to needs at the SDK layer, and `gmsol-utils` is one of the contract layer's dependencies. So if `strum` is really necessary, it should be put behind a feature as much as possible, for example, gated by the `display` feature.

angelo-75 (changes requested):
Sorry, I noticed this point when reviewing the implementation plan but couldn't come up with a good reason to reject it at the time. After rereading the related code, I now have a different opinion: in fact, the previous handling of flags has been fairly flat, that is, the bitmap gets expanded, so that downstream consumers don't need to repeat the unintuitive bitmap handling themselves, and this is also consistent with how the other serialization is done. As you said in the plan, this means we lose compatibility when new flags are added in the future, but I think it's worth it, and market statuses are unlikely to grow.

Therefore, I suggest changing this to a flat mapping of `bool` type `allow_*` fields (possibly wrapped in a `SerdeMarketStatusFlags`). This would change the shape of the whole implementation. What do you think?

angelo-75:
If you agree with this point, I will record this convention in the repository docs, so that we can avoid repeating this discussion in the future.

crates/utils/src/price/market_status.rs
gmitsj: Agreed - if we were continuing with this, we should gate it behind the display feature. However, since we have changed the architecture now, this is no longer necessary.

gmitsj:
Updated the implementation to use `SerdeMarketStatusFlags`, a flat mapping of `bool` type `allow_*` fields.

crates/cli/src/commands/market.rs
```diff
@@ -1692,6 +1695,155 @@ struct TokenMetadata {
...
+
+    if names.is_empty() {
+        "-".to_string()
+    } else {
+        names.join(", ")
+    }
+}
+
+#[cfg(test)]
+mod market_status_display_tests {
```
angelo-75: nit (non-blocking): Typically we'd go with a single `mod tests` instead of per-section test modules like `mod market_status_display_tests`. Worth considering a change, but this doesn't block the merge.

crates/cli/src/commands/market.rs
```diff
@@ -1692,6 +1695,155 @@ struct TokenMetadata {
...
+
+    #[test]
+    fn token_config_display_summarizes_each_provider() {
+        let mut feeds = IndexMap::new();
+        feeds.insert(
+            PriceProviderKind::Pyth,
+            feed_config(flags(false, true, true, false, false, false)),
+        );
+        feeds.insert(
+            PriceProviderKind::Switchboard,
```
angelo-75, line 1861: suggestion (non-blocking): Switchboard's encoding is not hex. Suggest adjusting it for consistency, but this doesn't block the merge.

crates/cli/src/commands/market.rs
```diff
@@ -1692,6 +1695,155 @@ struct TokenMetadata {
     init: bool,
 }
 
+/// [`SerdeTokenConfig`] augmented with a joined, human-readable summary of
+/// each feed's market-status flags, for CLI display only.
+#[derive(serde::Serialize)]
+struct TokenConfigDisplay {
+    #[serde(flatten)]
+    config: SerdeTokenConfig,
+    market_status: IndexMap<PriceProviderKind, String>,
```
angelo-75: suggestion (non-blocking): Consider renaming the display field to `market_status_summary`. It disambiguates from the structured `feeds.*.market_status` and matches the helper that produces it.

angelo-75 (dismissed):

crates/cli/src/commands/market.rs
```diff
@@ -443,13 +443,16 @@ impl super::Command for Market {
...
+                                ("market_status.chainlink_data_streams", "Chainlink Status",),
                                 ("feeds.pyth.feed_id", "Pyth Feed"),
                                 ("feeds.pyth.timestamp_adjustment", "Pyth TS Adj",),
+                                ("market_status.pyth", "Pyth Status"),
                                 ("feeds.switchboard.feed_id", "Switchboard Feed"),
                                 (
                                     "feeds.switchboard.timestamp_adjustment",
                                     "Switchboard TS Adj",
                                 ),
+                                ("market_status.switchboard", "Switchboard Status",),
```
angelo-75: **issue (blocking):** The table projection keys should be renamed accordingly, from `market_status.<provider>` to `market_status_summary.<provider>`. They still reference the old field name, so the Status columns render empty in table output. You can verify with:

```
cargo run -p gmsol-cli --features devnet -- -u devnet market tokens XaurYjNVXW8w61bXjWHGLJbGjcDhRhGDUwS1hEoLCp7
```

XAU has market-status flags configured on devnet, so the "Chainlink Status" row should show `allow_unknown, allow_pre_market, allow_post_market, allow_overnight, allow_closed` instead of an empty cell.

angelo-75 (changes requested):

crates/cli/src/commands/market.rs
gmitsj: Thanks for catching that! Updated below

angelo-75 (approved):

PR #393, opened 2026-07-30
feat: reject market-status flags unsupported by the configured feed

gmitsj:
## Summary

`set_feed_config_market_status_flag` previously let a `MARKET_KEEPER` enable any `MarketStatusFlag` on any feed, even when the feed could never actually report that status (e.g. a Pyth feed, or a Chainlink Data Streams feed whose report schema doesn't carry a given phase). The flag would silently persist with no effect, which is confusing operationally and easy to mis-verify.

This PR makes the program reject flags that the configured feed cannot support, adds an explicit opt-in (`allow_pending`) for staging such flags ahead of a feed upgrade, and surfaces the same check client-side (CLI/SDK) so operators get fast, pre-flight feedback instead of a failed transaction.

### `programs(store)`: reject unsupported flags

- `SetFeedConfigMarketStatusFlag::invoke_unchecked` now validates, when `enable` is `true`, that the flag is supported by the feed's configured provider/report schema. Unsupported flags are rejected with the new `CoreError::ProviderDoesNotSupportMarketStatus` (code 6044).
- Added an `allow_pending: bool` instruction argument. When set, an unsupported flag is accepted anyway and stored as pending, pre-staged policy with no effect until the feed can support it (e.g. after a schema upgrade). Disabling a flag never requires `allow_pending`.
- `chainlink-datastreams(gmsol)`: added `is_market_status_flag_supported`, which decodes the report schema version from a Chainlink feed id and determines which `MarketStatusFlag`s that schema can ever emit:
  - v2/v3/v7: no market status at all — no flag supported.
  - v4/v8: only `Unknown`/`RegularHours`/`Closed` — phase-specific flags (`AllowPreMarket`, `AllowPostMarket`, `AllowOvernight`) not supported.
  - v11: all phases — every flag treated as potentially effective.
  - any other/unrecognized version: treated conservatively as unsupported.
- Non-Chainlink providers never support any market-status flag.
- Extended `tests/anchor_test/market_status.rs` to cover: rejection for non-status providers, rejection for schema versions that don't carry a given phase, acceptance via `allow_pending`, and full acceptance on a v11 schema feed.

### `sdk(cli)`: fail fast before sending

- `gmsol_sdk::ops::token_config::is_market_status_flag_supported` (behind the `chainlink` feature) mirrors the on-chain check so callers can pre-validate without a round trip.
- CLI `market set-market-status-flag` now checks `--enable` flags against each token's configured feed before building any transaction, and bails out with a clear message (naming the offending token/flags) if any are unsupported — pointing the user at the new `--allow-pending` flag instead of letting the transaction fail on-chain.

### `sdk/cli`: surface effective vs. pending status

- Added `SerdeMarketStatusSupport` to `SerdeFeedConfig`, reporting per-flag whether it is currently effective (`Some(true)`), stored as inert pending policy (`Some(false)`), or undeterminable in this build (`None` — e.g. a Chainlink feed inspected without the `chainlink` feature).
- `cli market tokens` now annotates each enabled flag in its summary, e.g. `allow_unknown (pending)` or `allow_unknown (support unknown)`, instead of just listing enabled flag names.

## Breaking changes

- `set_feed_config_market_status_flag` gains a new `allow_pending: bool` instruction argument (also threaded through `TokenConfigOps::set_feed_config_market_status_flag` in the Rust SDK and the CLI). Existing callers must pass a value — `false` preserves the previous "reject silently-ineffective flags" intent, though note the *previous* behavior was to accept and silently no-op, not reject.
- New `CoreError` variant `ProviderDoesNotSupportMarketStatus` (code 6044) shifts all subsequent `CoreError` codes up by one (IDL regenerated accordingly).

## Testing

- `tests/anchor_test/market_status.rs`: extended existing test plus a new `set_feed_config_market_status_flag_schema_aware` test covering v8/v11 synthetic Chainlink feed schemas.
- `crates/cli/src/commands/market.rs`: unit tests for the market-status summary formatting (effective / pending / support-unknown annotations).

### DIFF
--- a/programs/store/src/lib.rs
+++ b/programs/store/src/lib.rs
@@ -892,12 +892,20 @@ pub mod gmsol_store {
     ///   Must be a valid [`PriceProviderKind`] value.
     /// - `flag`: The [`MarketStatusFlag`] index to set.
     /// - `enable`: Enable or disable the flag.
+    /// - `allow_pending`: When `enable` is `true` and the `provider` is not
+    ///   [`ChainlinkDataStreams`](PriceProviderKind::ChainlinkDataStreams), the flag
+    ///   is stored as pending, pre-staged policy that has no effect while the
+    ///   provider does not report market status, instead of being rejected.
+    ///   Ignored otherwise.
     ///
     /// # Errors
     /// - The [`authority`](SetFeedConfigMarketStatusFlag::authority) must be a
     ///   signer and a MARKET_KEEPER in the given store.
     /// - The [`token`](SetFeedConfigMarketStatusFlag::token) must exist in the token map.
     /// - The `provider` index must correspond to a valid [`PriceProviderKind`].
+    /// - When `enable` is `true`, the `provider` must be
+    ///   [`ChainlinkDataStreams`](PriceProviderKind::ChainlinkDataStreams) unless
+    ///   `allow_pending` is set, since other providers do not report market status.
     /// - The feed config of the `provider` for the `token` must be initialized.
     /// - `flag` must be a valid [`MarketStatusFlag`] value.
     #[access_control(internal::Authenticate::only_market_keeper(&ctx))]
@@ -906,6 +914,7 @@ pub mod gmsol_store {
         provider: u8,
         flag: u8,
         enable: bool,
+        allow_pending: bool,
     ) -> Result<()> {
         let token = ctx.accounts.token.key();
         let authorized =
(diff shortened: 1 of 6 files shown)


programs/store/src/lib.rs
```diff
@@ -4078,6 +4084,9 @@ pub enum CoreError {
     /// Not supported price provider for custom price feed.
     #[msg("this price provider is not supported to be used with custom price feed")]
     NotSupportedCustomPriceProvider,
+    /// Provider does not support market status.
+    #[msg("this price provider does not support market status")]
+    ProviderDoesNotSupportMarketStatus,
```
angelo-75: issue (blocking): Error variants must be append-only. Inserting `ProviderDoesNotSupportMarketStatus` at 6044 shifts all subsequent `CoreError` codes by one, breaking every external consumer that matches on numeric codes. Please append it to the end of the enum instead and regenerate the IDL.

angelo-75 (changes requested):
issue (blocking): The approach taken expands the scope too much. For now, we only require blocking the configuration of unsupported providers. Fine-grained market status flag configuration is not required (v4 will be removed, and v8 does not produce extra market statuses and may also have its support removed in the future). allow_pending is not needed either (adding it brings in too many accompanying changes, while the feature will most likely never be used).

programs/store/src/lib.rs
gmitsj: Created a new section called 
```  
    // ===========================================
    //              Oracle Errors (2)
    // ===========================================
```
    
and appended to the bottom of the enum.

PR #405, opened 2026-08-11
feat(store): initialize increase orders' final output token at creation

crypto1gmt:
## Summary

An increase order can now initialize its final output token at creation. The escrow stays **optional**
for increase orders, exactly as before: when it is passed, both halves of the order's final output token
slot are recorded and the mint is validated to be the position's collateral token; when it is omitted,
the slot stays uninitialized and nothing changes.

This is what makes an increase order eligible for a builder fee later. The fee is charged in the final
output token and paid out of that escrow (ADR-0001), so an order created without the escrow can never be
given one. Requiring the escrow to be present is deliberately not done here; it belongs to the
owner-signed `set_builder_fee` instruction.

## What's here

- `create_order_v2` validates `final_output_token == collateral_token` for increase orders before it
  builds the operation. That account is always provided, so the check covers every increase-creation path,
  and it keeps `init_increase` free of a token concern that means nothing at that level. It reuses
  `CoreError::TokenMintMismatched`, the same error the neighbouring long/short token checks raise, instead
  of adding a new error variant.
- `CreateIncreaseOrderOperation` takes the optional escrow as its source of truth. When it is provided, its
  mint is validated against the collateral token and `tokens.final_output_token` is initialized from it;
  when it is omitted the slot stays uninitialized. This is the second of the two layers, and the only check
  that can run there, since without the escrow the operation never learns which mint was requested.
- `execute_increase_position` validates that the stored final output token equals the position's collateral
  token, **only when that slot is set**. Orders created before this change carry `None` and are skipped,
  otherwise every one of them would stop executing.
- Doc comments: `final_output_token_escrow` is now documented as optional for increase orders rather than
  "only required by decrease and swap orders", and both validation sites carry a comment explaining what a
  builder fee will require of them.
- SDK: `CreateOrderBuilder::prepare_final_output_token_escrow`, off by default (see "For review").

## Behavior change

One, and it is the intended one: creating an increase order whose final output token differs from the
collateral token used to succeed and silently ignore the value. It now reverts with `TokenMintMismatched`
and no order is created. Everything else is additive.

Failure semantics at execution time, since the ACs do not spell them out: an error raised inside
`perform_execution` is turned into a cancellation unless the keeper passes `throw_on_execution_error`, so a
mismatch on an already-created order cancels that order and the keeper's transaction still succeeds. This
matches ADR-0001's "the execution-side rejection cancels the order rather than reverting".

## Breaking-change verification

Checked every surface the optional-escrow shape touches. **Verdict: non-breaking on all of them, no client
or keeper release coordination needed.**

| # | Surface | Finding |
|---|---|---|
| 1 | Create, old client (passes `None`) | Identical to today: the slot stays uninitialized. Grandfathered by design; such orders simply cannot receive a builder fee later. |
| 2 | Create, new client + old program | Safe in either deploy order: `final_output_token_escrow` is already `Option` in `CreateOrderV2` with only ATA constraints, and today's program validates then ignores it on increase. |
| 3 | Create, `Some` + new program | The ATA constraints already force `escrow == ATA(order, final_output_token)`; this PR adds the mint check and the init. Since the mint must equal the collateral token, the address coincides with the collateral-side escrow, so the stored bytes match the shape decrease orders already produce today. |
| 4 | Execute, on-chain | The constraint is `Option` equality, stored vs passed. Old orders (`None`) need `None`: unchanged. New orders (`Some`) need `Some`: see #5. |
| 5 | Execute, keeper side | The SDK's increase/swap branch maps `final_output_token_and_account` read from the stored order as a plain `Option`, with no order-kind gating and no `ok_or`, so even an old keeper binary auto-includes the mint, escrow and vault for new orders. The `ok_or("missing final output token")` path exists only in the decrease branch, where the hint is always `Some`. |
| 6 | Execute, runtime | No builder fee exists yet, so `TransferOut.final_output_token` stays 0 for increase: the accounts are present but untouched. |
| 7 | Close, on-chain | Constraint is again stored vs passed. `transfer_to_atas` keeps a `seen` set and skips an escrow it has already handled, so the duplicated final == collateral-side escrow is transferred and closed exactly once by explicit design. The payout target is the receiver's ATA of the same mint either way. |
| 8 | Close, SDK side | Same plain-`Option` hint mapping as #5, no kind gating: old binaries adapt on their own. |
| 9 | `update_order_v2` | Touches params only, cannot corrupt the initialized token slots. |
| 10 | IDL / ABI | No account-list or argument changes; the account is already optional in the IDL, only doc comments move. Passing the escrow costs about one byte on the wire, since it is the same pubkey as the long/short escrow and message-level account dedup applies. |
| 11 | Order-kind inference | Grepped `programs/` and `crates/` for presence-based inference on the final output token: nothing treats "has a final output token" as "is a decrease or swap order". |

One caveat worth stating as a client contract: the above holds for clients that assemble accounts from the
order's stored token hints, which is what the SDK does on every v2 path. A third-party integrator that
hardcodes `None` for the increase final-output accounts instead of reading the hints would fail the
stored-vs-passed constraint on new orders.

## Tests

Three cases in `tests/anchor_test/order.rs`:

1. Without the escrow, the final output token slot stays uninitialized while the long and short slots are
   filled. This is the backward-compatibility case and the default every client hits today.
2. With the escrow, both halves are set to the collateral token and its order-owned ATA, and the order is
   then executed for real, so the new execution-time check runs on a live path rather than being asserted
   about.
3. A mismatching final output token reverts with `TokenMintMismatched` and leaves no order account.

The third one has a wrinkle: with the escrow present a wrong mint is unreachable, because the
`associated_token::mint = final_output_token` constraint on the escrow rejects it before either
validation runs. The test therefore builds the instruction without the escrow, which is both the only
path where the mint is unconstrained and the only case the instruction-layer check alone can catch. The
upside is that with the escrow present, two independent layers enforce the same invariant.

## For review

- The SDK opt-in is not covered by the ACs. Today the SDK prepares the final-output escrow only for swap
  and decrease orders, so no client could create an increase order carrying it, and therefore none could ever
  be given a builder fee. I added the minimal opt-in here, off by default, because the tests need it anyway.
  Happy to pull it out if you would rather it landed with the fee work.
- This rests on one assumption: the builder fee is charged in the final output token for increase as well as
  decrease. The creation-time mint validation is what would need revisiting if that direction changes.
- `builders/order/create.rs` has the same integration gap and is deliberately not touched here. Nothing is
  broken there today, since `receive_token` is `None` for increase orders and the mint therefore already
  resolves to the collateral token; what is missing is the opt-in to pass the escrow. Follow-up PR right
  after this one.

## Out of scope

- Requiring the escrow to be present. That check belongs to `set_builder_fee`.
- Any fee computation, accrual or routing.
- Swap and decrease paths: untouched, in both creation and execution.

### DIFF
--- a/programs/store/src/instructions/exchange/order.rs
+++ b/programs/store/src/instructions/exchange/order.rs
@@ -405,6 +405,22 @@ impl<'info> internal::Create<'info, Order> for CreateOrderV2<'info> {
                     .execute()?;
             }
             OrderKind::MarketIncrease | OrderKind::LimitIncrease => {
+                // The final output token is where a builder fee would be charged, so for an
+                // increase order it must be the position's collateral token. The check lives here
+                // rather than in the operation because the escrow is optional: without it the ops
+                // layer never sees the requested mint, while this account is always provided, so
+                // this is the only place that covers every increase-creation path.
+                let collateral_token = if params.is_collateral_long {
+                    self.market.load()?.meta().long_token_mint
+                } else {
+                    self.market.load()?.meta().short_token_mint
+                };
+                require_keys_eq!(
+                    self.final_output_token.key(),
+                    collateral_token,
+                    CoreError::TokenMintMismatched
+                );
+
                 let initial_collateral = self
                     .initial_collateral_token_escrow
                     .as_ref()
@@ -426,8 +442,7 @@ impl<'info> internal::Create<'info, Order> for CreateOrderV2<'info> {
                     .initial_collateral_token(initial_collateral.as_ref())
                     .long_token(long_token.as_ref())
                     .short_token(short_token.as_ref())
-                    .final_output_token(&self.final_output_token)
-                    .final_output_token_escrow(self.final_output_token_escrow.as_deref())
+                    .final_output_token(self.final_output_token_escrow.as_deref())
                     .build()
                     .execute()?;
             }
--- a/programs/store/src/instructions/exchange/order.rs
+++ b/programs/store/src/instructions/exchange/order.rs
@@ -259,7 +259,11 @@ pub struct CreateOrderV2<'info> {
     )]
     pub initial_collateral_token_escrow: Option<Box<Account<'info, TokenAccount>>>,
     /// Final output token escrow account.
-    /// Only required by decrease and swap orders.
+    ///
+    /// Required by decrease and swap orders. Optional for increase orders: when provided, it is
+    /// recorded on the order and becomes the account a builder fee would be paid out of, and when
+    /// omitted the order's final output token is left uninitialized, so no builder fee can be set
+    /// on it later.
     #[account(
         mut,
         associated_token::mint = final_output_token,
@@ -422,6 +426,8 @@ impl<'info> internal::Create<'info, Order> for CreateOrderV2<'info> {
                     .initial_collateral_token(initial_collateral.as_ref())
                     .long_token(long_token.as_ref())
                     .short_token(short_token.as_ref())
+                    .final_output_token(&self.final_output_token)
+                    .final_output_token_escrow(self.final_output_token_escrow.as_deref())
                     .build()
                     .execute()?;
             }
--- a/programs/store/src/ops/order.rs
+++ b/programs/store/src/ops/order.rs
@@ -149,7 +149,7 @@ impl<'a, 'info> CreateOrderOperation<'a, 'info> {
     ) -> CreateIncreaseOrderOperationBuilder<
         'a,
         'info,
-        ((CreateOrderOperation<'a, 'info>,), (), (), (), ()),
+        ((CreateOrderOperation<'a, 'info>,), (), (), (), (), (), ()),
     > {
         CreateIncreaseOrderOperation::builder().common(self)
     }
@@ -356,6 +356,20 @@ pub(crate) struct CreateIncreaseOrderOperation<'a, 'info> {
     initial_collateral_token: &'a Account<'info, TokenAccount>,
     long_token: &'a Account<'info, TokenAccount>,
     short_token: &'a Account<'info, TokenAccount>,
+    /// The final output token mint, always provided by the instruction.
+    ///
+    /// It is validated to be the collateral token in
+    /// [`init_increase`](OrderActionParams::init_increase).
+    final_output_token: &'a Account<'info, Mint>,
+    /// The escrow for the final output token.
+    ///
+    /// Optional for increase orders: clients that do not intend to use a builder fee can keep
+    /// omitting it, and the order's final output token is then left uninitialized, exactly as
+    /// before this account was read at all. When it is provided, both halves of
+    /// [`OrderTokenAccounts::final_output_token`] are initialized from it. The account is
+    /// constrained to be the ATA of `final_output_token` owned by the order, so no further
+    /// validation of the escrow itself is needed here.
+    final_output_token_escrow: Option<&'a Account<'info, TokenAccount>>,
 }
 
 impl CreateIncreaseOrderOperation<'_, '_> {
@@ -376,11 +390,15 @@ impl CreateIncreaseOrderOperation<'_, '_> {
                     .init(self.initial_collateral_token);
                 tokens.long_token.init(self.long_token);
                 tokens.short_token.init(self.short_token);
+                if let Some(escrow) = self.final_output_token_escrow {
+                    tokens.final_output_token.init(escrow);
+                }
                 params.init_increase(
                     create.is_long,
                     create.kind,
                     self.position.key(),
                     collateral_token,
+                    self.final_output_token.key(),
                     create.initial_collateral_delta_amount,
                     create.size_delta_value,
                     create.trigger_price,
@@ -1499,6 +1517,26 @@ fn execute_increase_position(
     event: &mut TradeData,
     order: &mut Order,
 ) -> Result<u128> {
+    // The builder fee is charged in the order's final output token, so for an increase order that
+    // token, and therefore the escrow it is paid out of, must belong to the position's collateral
+    // token. Orders created before the final output token was initialized at creation time carry
+    // `None` here and are skipped, otherwise every one of them would stop executing.
+    //
+    // The escrow itself needs no check: the account passed to the execute instruction is
+    // constrained to equal the one stored on the order, and to be the ATA of the passed final
+    // output token mint. Since an ATA address determines its mint, a mismatching mint cannot get
+    // this far.
+    //
+    // Once the builder fee lands, `set_builder_fee` is what requires the escrow to be present, so
+    // that is what makes this validation run for every fee-bearing order.
+    if let Some(final_output_token) = order.tokens.final_output_token.token() {
+        require_keys_eq!(
+            final_output_token,
+            *position.collateral_token(),
+            CoreError::TokenMintMismatched
+        );
+    }
+
     let params = &order.params;
 
     // Perform swap.
--- a/programs/store/src/ops/order.rs
+++ b/programs/store/src/ops/order.rs
@@ -149,7 +149,7 @@ impl<'a, 'info> CreateOrderOperation<'a, 'info> {
     ) -> CreateIncreaseOrderOperationBuilder<
         'a,
         'info,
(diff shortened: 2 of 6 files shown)


programs/store/src/ops/order.rs
```diff
@@ -356,6 +356,20 @@ pub(crate) struct CreateIncreaseOrderOperation<'a, 'info> {
     initial_collateral_token: &'a Account<'info, TokenAccount>,
     long_token: &'a Account<'info, TokenAccount>,
     short_token: &'a Account<'info, TokenAccount>,
+    /// The final output token mint, always provided by the instruction.
+    ///
+    /// It is validated to be the collateral token in
+    /// [`init_increase`](OrderActionParams::init_increase).
+    final_output_token: &'a Account<'info, Mint>,
```
angelo-75: issue (blocking): This should follow the same convention as the other token accounts: use the actually provided `TokenAccount` as the source of truth, rather than passing in an extra final output token mint. This reduces the assumptions the ops layer makes about tokens. So this field should be removed.

programs/store/src/ops/order.rs
```diff
@@ -356,6 +356,20 @@ pub(crate) struct CreateIncreaseOrderOperation<'a, 'info> {
...
+    /// The escrow for the final output token.
+    ///
+    /// Optional for increase orders: clients that do not intend to use a builder fee can keep
+    /// omitting it, and the order's final output token is then left uninitialized, exactly as
+    /// before this account was read at all. When it is provided, both halves of
+    /// [`OrderTokenAccounts::final_output_token`] are initialized from it. The account is
+    /// constrained to be the ATA of `final_output_token` owned by the order, so no further
+    /// validation of the escrow itself is needed here.
+    final_output_token_escrow: Option<&'a Account<'info, TokenAccount>>,
```
angelo-75: issue (blocking): paired with the previous comment, this should be renamed to final_output_token for naming consistency.

programs/store/src/states/order.rs
```diff
@@ -751,13 +751,23 @@ impl OrderActionParams {
...
         valid_from_ts: Option<i64>,
     ) -> Result<()> {
+        // The builder fee is charged in the order's final output token, so for an increase order
+        // that token has to be the position's collateral token. This is the choke point every
+        // increase-order creation path goes through, so validating here covers all of them.
+        require_keys_eq!(
+            final_output_token,
+            collateral_token,
+            CoreError::TokenMintMismatched
+        );
```
angelo-75: nit: if we go with the approach of only passing in the final output token account, all we can get here is an optional final output token, so the validation can be changed to run only in the Some case. This is acceptable for the ops layer. To keep the instruction consistent, the instruction layer can additionally validate that the final output token == the collateral token (this kind of duplicated validation is worthwhile: compared with the slight waste of CU, layered verifiability and defense in depth in contract code are far more valuable).

angelo-75 (changes requested):
Besides the issues related to the code changes, there are also the following:

* suggestion: since at the ops layer we plan to validate final output token == collateral token only optionally, the instruction layer should add an extra validation to satisfy the convention stated in the docs (or the docs could be updated instead, but I think adding the validation is more reasonable, as there is currently no reason to relax this constraint).
* issue (non-blocking): the SDK changes in this PR only cover the Client part, but there is actually another integration in crates/sdk/src/builders/order/create.rs that also needs to be updated. This can be done in a separate PR though. It is really a legacy issue: we should migrate the old Client-based instruction integration to the atomic group based one (which is the newer design).
* nit (non-blocking): the Changelog should point out that these additional validations are breaking changes, but we usually re-confirm the Changelog before release anyway.
* suggestion (non-blocking): since the instruction docs have changed, the IDL needs to be updated accordingly, but we usually regenerate it before release or when necessary, to reduce code conflicts.

After the adjustments are made as suggested, the final output token invariant of an Increase Order becomes: final output token escrow == ATA(order, collateral token), or final output token == final output token escrow == DEFAULT_TOKEN.

programs/store/src/states/order.rs
angelo-75: In addition, I'm also reconsidering our decision about where the validation should live. I now think the validation should happen before the call to init_increase, rather than inside init_increase. The reason is that final_output_token has no real meaning from init_increase's point of view. If we go this way, we can avoid any modification to init_increase (no new final_output_token parameter is introduced), and the collateral_token == final_output_token validation can be done within the same if let Some() block.

What do you think?

crypto1gmt: Agreed, and I think it's the better split. `init_increase` is about the order's params, and a final output token doesn't mean anything at that level, so passing one in just to validate it there was the wrong shape. Moving it up also removes the parameter, so `init_increase` stays untouched in this PR.

It lands cleanly too: `collateral_token` is already computed just above the `init_with` closure, so the check fits into the same `if let Some(...)` block with nothing extra to thread through.

One consequence of that. Once the ops-layer check sits inside `if let Some(...)`, "every increase creation path is validated" rests entirely on the instruction-layer check. That holds today because `create_order_v2` is the only thing that builds `CreateIncreaseOrderOperation`, but it becomes a load-bearing assumption rather than an incidental one, so any future creation instruction has to carry the same check.

Also, since the escrow is constrained to `ATA(order, final_output_token)`, `escrow.mint == final_output_token` is already guaranteed, and the ops-layer check follows from the instruction-layer one. Keeping it anyway for the reason you gave, just noting it's a deliberate second line rather than an accidental duplicate.

One thing that needs your call: AC3 currently pins the validation to `init_increase` as the choke point. If we go this way, the AC text stops matching the code. Do you want to update it, or should I record the deviation in the PR description?

programs/store/src/ops/order.rs
crypto1gmt: Agreed, removing the field. The mint comes off the escrow account itself, so there's nothing extra to pass in.

programs/store/src/ops/order.rs
crypto1gmt: Will rename it.

crypto1gmt:
The IDL: this PR doesn't touch `crates/programs/idls/` at all, so there's nothing to undo. Happy to leave it to the release regeneration.

I'll take `builders/order/create.rs`, but in a follow-up PR right after this one rather than growing this diff. Nothing is broken there today: `receive_token` is `None` for increase orders, so `final_output_token` already resolves to the collateral token and the new validation passes unchanged. What's missing is the opt-in to pass the escrow, without which an order created through that builder can never become fee-eligible.

Changelog: I'll mark the added validations as breaking.

programs/store/src/ops/order.rs
crypto1gmt: Done in `c74dc491`. The field is gone and the mint is read off the escrow account.

programs/store/src/ops/order.rs
crypto1gmt: Done in `c74dc491`.

programs/store/src/states/order.rs
crypto1gmt: Implemented in `c74dc491`. `states/order.rs` is back to untouched, the ops-layer check runs inside the `if let Some(...)` block, and `create_order_v2` carries the unconditional one. I updated the PR description too, since it still described the old placement.

programs/store/src/instructions/exchange/order.rs
```diff
@@ -401,6 +405,22 @@ impl<'info> internal::Create<'info, Order> for CreateOrderV2<'info> {
...
+                // The final output token is where a builder fee would be charged, so for an
+                // increase order it must be the position's collateral token. The check lives here
+                // rather than in the operation because the escrow is optional: without it the ops
+                // layer never sees the requested mint, while this account is always provided, so
+                // this is the only place that covers every increase-creation path.
+                let collateral_token = if params.is_collateral_long {
+                    self.market.load()?.meta().long_token_mint
+                } else {
+                    self.market.load()?.meta().short_token_mint
+                };
```
angelo-75, line 417: nit (non-blocking): `CreateOrderParams.collateral_token()` could actually be used here, but we don't need to deal with it this time.

angelo-75 (approved):

programs/store/src/instructions/exchange/order.rs
angelo-75: Tracked internally.

PR #406, opened 2026-08-12
feat: land builder fee computing and routing behind a zero factor

gmitsj:
CON-34 Land builder fee computation and routing behind a zero factor

## Summary

This is the productionized implementation of builder fee computation for the increase and decrease execution paths, per ADR-0001 and the follow-up issue that supersedes PoC #401. The fee is charged from collateral at execution time, denominated in each order's final output token (the collateral token for increase, the post-swap receive token for decrease). Everything ships inert. Every real call site still hardcodes a builder fee factor of 0, so production behavior is unchanged until a later activation issue replaces that constant with the order's snapshotted factor.

## What's implemented

- `compute_builder_fee_amount` / `clamp_builder_fee_amount`. Pure USD-value to token-amount conversion (`apply_factor` against `size_delta_usd`, rounded up against the min price) and shortfall clamping.
- Increase orders now initialize `tokens.final_output_token` at creation, required to equal the collateral token. This field was previously left unset for increase orders even though the account was already threaded through `create_order_v2`. `final_output_token_escrow` is now a required account for increase-order creation, so the Rust SDK's order-creation builder was updated to supply it.
- `charge_builder_fee_on_collateral_increment`. Computes and charges the fee in the collateral token, after the pay token has been swapped into it. Underpayment is not tolerated here. A shortfall returns `CoreError::BuilderFeeExceedsCollateral`, cancelling the order rather than charging a partial amount.
- `execute_increase_position` wiring. Validates `final_output_token == collateral_token` (a backward-compatibility guard, `CoreError::BuilderFeeFinalOutputTokenMismatch`, for orders created before the field was initialized), charges the fee post-swap, routes it through `transfer_out`'s final-output-token bucket, and emits `BuilderFeeCharged`.
- `estimate_builder_fee_for_collateral_withdrawal`. Folds a fee estimate into a decrease order's collateral withdrawal amount, so it stays fundable even when the position closes at a loss. Rejects `DecreasePositionSwapType::CollateralToPnlToken` whenever the fee is non-zero, via `CoreError::BuilderFeeSwapTypeNotAllowed`.
- `execute_decrease_position` wiring. The actual charge happens after the receive-token swap, against the final output token (which may differ from the collateral token), computed fresh against that token's price. Underpayment is tolerated here, charging whatever is available. Emits `BuilderFeeCharged`.
- `BuilderFeeCharged` CPI event. Records both the payable (pre-clamp) and paid (post-clamp) amounts. Not emitted when the factor is 0. Emitted even when the paid amount is 0, as long as the factor is non-zero, so indexers can tell "no builder attached" apart from "builder attached, nothing collectible."

## Why the fee token changed from the PoC

PoC #401 charged the increase fee in the pay token before the swap, and the decrease fee in the collateral token before the receive-token swap. Review settled on a different, cheaper design instead. Charge both in the order's final output token, after any swap.

For increase, the final output token is now required to equal the collateral token, so the fee computation reuses the collateral price that's already loaded for the instruction instead of fetching a separate price for an arbitrary pay token.

For decrease, the final output token can legitimately differ from the collateral token. This is what keeps the design compatible with a future cross-collateral feature, where the collateral token might not be the long or short token at all. The `CollateralToPnlToken` rejection is still required, but for a different reason than before. That swap type deterministically routes the entire collateral-token output into the pnl token before the receive-token swap even runs, so the final output token bucket would be empty on every single order that picks it. That's not ordinary underpayment variance, which decrease already tolerates via `NoSwap`. It's a guaranteed way to pay no fee at all.

## Acceptance criteria

- [x] Builder fee charged from the final output token on both paths, counted into the final output token output.
- [x] Accrual into `order.builder_fee_amount` is out of scope. TODO comments mark both accrual sites.
- [x] Fee computed as `apply_factor(size_delta_usd, factor)`, converted using the min price, rounded up.
- [x] Decrease underpayment is not restricted. Charges whatever is available and completes.
- [x] Increase underpayment cancels the order.
- [x] Increase deduction happens after the swap and before `position.increase()`, so the fee never enters any market's balances. Commented at the deduction site.
- [x] Decrease deduction happens after the swap, naturally netted out of the market's balances. Commented at the deduction site.
- [x] `CollateralToPnlToken` rejected on decrease when the factor is non-zero.
- [x] `BuilderFeeCharged` records both amounts, gated on the factor rather than the paid amount, emitted via `event_emitter`.
- [x] Every real call site hardcodes factor 0. No existing instruction's behavior changes.

## Out of scope

- Account and schema changes (`Order.builder_fee_amount`, `UserHeader.builder_fee_factor`, `Store.max_builder_fee_factor`, the feature flag, `UserTokenController`). Covered by a separate, independent issue.
- The order-creation snapshot instruction that would populate a real, non-zero factor.
- Settlement, claim, and permissionless configuration.

## Testing

- `cargo test -p gmsol-store`. 22 passed, 0 failed.
- `cargo build` / `cargo check --workspace` / `cargo clippy --workspace --lib`. Clean, no new warnings.
- No behavioral change to any existing on-chain path. Verified by inspection that every real caller passes `builder_fee_factor = 0`, which short-circuits before touching any price or account.

### DIFF
--- a/programs/store/src/lib.rs
+++ b/programs/store/src/lib.rs
@@ -4438,6 +4438,12 @@ pub enum CoreError {
     /// order instead of charging a partial fee.
     #[msg("builder fee exceeds the collateral increment")]
     BuilderFeeExceedsCollateral,
+    /// A non-zero builder fee is being charged on an increase order that
+    /// has no final output token. Initializing it is optional for
+    /// increase orders, and the builder fee is charged in it, so an order
+    /// that never opted in cannot pay one.
+    #[msg("builder fee requires the final output token to equal the collateral token")]
+    BuilderFeeFinalOutputTokenMismatch,
     // NOTE: New variants must be appended here to keep existing error codes stable.
 }
 
--- a/programs/store/src/lib.rs
+++ b/programs/store/src/lib.rs
@@ -4444,6 +4444,14 @@ pub enum CoreError {
     /// that never opted in cannot pay one.
     #[msg("builder fee requires the final output token to equal the collateral token")]
     BuilderFeeFinalOutputTokenMismatch,
+    /// A decrease order with a non-zero builder fee used
+    /// `DecreasePositionSwapType::CollateralToPnlToken`. That swap type
+    /// deterministically moves the entire collateral-token output into the
+    /// pnl token before the receive-token swap runs, leaving the final
+    /// output token bucket at zero every time, which would make the
+    /// builder fee trivially and reliably bypassable.
+    #[msg("this decrease position swap type is not allowed together with a builder fee")]
+    BuilderFeeSwapTypeNotAllowed,
     // NOTE: New variants must be appended here to keep existing error codes stable.
 }
 
--- a/programs/store/src/lib.rs
+++ b/programs/store/src/lib.rs
@@ -4432,6 +4432,12 @@ pub enum CoreError {
     /// fee; settle it first via `settle_builder_fee`.
     #[msg("order has an unsettled builder fee")]
     UnsettledBuilderFee,
+    /// An increase order's collateral increment cannot fully cover the
+    /// builder fee. Unlike decrease, increase has no equivalent of
+    /// "complete anyway at a loss" to fall back on, so this cancels the
+    /// order instead of charging a partial fee.
+    #[msg("builder fee exceeds the collateral increment")]
+    BuilderFeeExceedsCollateral,
     // NOTE: New variants must be appended here to keep existing error codes stable.
 }
 
--- a/programs/store/src/ops/order.rs
+++ b/programs/store/src/ops/order.rs
@@ -1544,6 +1544,39 @@ fn execute_swap(
     Ok(())
 }
 
+/// Charges the builder fee against an increase order's collateral
+/// increment, in the collateral token, after the pay token has already
+/// been swapped into it. Returns the reduced collateral increment
+/// together with the fee actually charged.
+///
+/// Unlike the decrease path, underpayment here is not tolerated: an
+/// increase order has no equivalent of "complete the close anyway, at a
+/// loss" to fall back on, so if the collateral increment can't fully
+/// cover the fee, this returns an error instead of charging a partial
+/// amount. The caller propagates this as a soft failure (the order is
+/// cancelled), not a transaction revert.
+fn charge_builder_fee_on_collateral_increment(
+    collateral_increment_amount: u64,
+    size_delta_usd: u128,
+    builder_fee_factor: u128,
+    collateral_price: &Price<u128>,
+) -> Result<(u64, u64)> {
+    let payable_amount =
+        compute_builder_fee_amount(size_delta_usd, builder_fee_factor, collateral_price)?;
+    let payable_amount_u64 =
+        u64::try_from(payable_amount).map_err(|_| error!(CoreError::TokenAmountOverflow))?;
+    require_gte!(
+        collateral_increment_amount,
+        payable_amount_u64,
+        CoreError::BuilderFeeExceedsCollateral
+    );
+    // The assertion above guarantees the subtraction succeeds.
+    let collateral_increment_after_fee = collateral_increment_amount
+        .checked_sub(payable_amount_u64)
+        .ok_or_else(|| error!(CoreError::TokenAmountOverflow))?;
+    Ok((collateral_increment_after_fee, payable_amount_u64))
+}
+
 #[inline(never)]
 fn execute_increase_position(
     oracle: &Oracle,
@@ -2136,5 +2169,77 @@ mod tests {
     #[test]
     fn builder_fee_within_available_is_unchanged() {
         assert_eq!(clamp_builder_fee_amount(10, 100), 10);
+    }
+
+    #[test]
+    fn zero_builder_fee_factor_leaves_collateral_increment_unchanged() {
+        let (collateral_increment_after_fee, fee) = charge_builder_fee_on_collateral_increment(
+            1_000,
+            100_000 * constants::MARKET_USD_UNIT,
+            0,
+            &price(
+                2 * constants::MARKET_USD_UNIT,
+                2 * constants::MARKET_USD_UNIT,
+            ),
+        )
+        .unwrap();
+        assert_eq!(fee, 0);
+        assert_eq!(collateral_increment_after_fee, 1_000);
+    }
+
+    #[test]
+    fn charges_expected_builder_fee_on_collateral_increment() {
+        // $100,000 size, 5 bps factor => $50 fee value, at $2/unit => 25 units.
+        let size_delta_usd = 100_000 * constants::MARKET_USD_UNIT;
+        let factor = constants::MARKET_USD_UNIT / 2_000;
+        let (collateral_increment_after_fee, fee) = charge_builder_fee_on_collateral_increment(
+            1_000,
+            size_delta_usd,
+            factor,
+            &price(
+                2 * constants::MARKET_USD_UNIT,
+                2 * constants::MARKET_USD_UNIT,
+            ),
+        )
+        .unwrap();
+        assert_eq!(fee, 25);
+        assert_eq!(collateral_increment_after_fee, 975);
+    }
+
+    #[test]
+    fn exact_builder_fee_consumes_the_whole_collateral_increment() {
+        // Fee equals the entire collateral increment: allowed, leaves 0.
+        let size_delta_usd = 100_000 * constants::MARKET_USD_UNIT;
+        let factor = constants::MARKET_USD_UNIT / 2_000;
+        let (collateral_increment_after_fee, fee) = charge_builder_fee_on_collateral_increment(
+            25,
+            size_delta_usd,
+            factor,
+            &price(
+                2 * constants::MARKET_USD_UNIT,
+                2 * constants::MARKET_USD_UNIT,
+            ),
+        )
+        .unwrap();
+        assert_eq!(fee, 25);
+        assert_eq!(collateral_increment_after_fee, 0);
+    }
+
+    #[test]
+    fn builder_fee_underpayment_errors_instead_of_clamping() {
+        // Same fee value as above (25 units), but the swap only produced 10 units.
+        let size_delta_usd = 100_000 * constants::MARKET_USD_UNIT;
+        let factor = constants::MARKET_USD_UNIT / 2_000;
+        let err = charge_builder_fee_on_collateral_increment(
+            10,
+            size_delta_usd,
+            factor,
+            &price(
+                2 * constants::MARKET_USD_UNIT,
+                2 * constants::MARKET_USD_UNIT,
+            ),
+        )
+        .unwrap_err();
+        assert_eq!(err, error!(CoreError::BuilderFeeExceedsCollateral));
     }
 }
--- a/programs/store/src/ops/order.rs
+++ b/programs/store/src/ops/order.rs
@@ -1740,6 +1740,60 @@ fn execute_increase_position(
     Ok(paid_order_fee_value)
 }
 
+/// Folds an estimate of the builder fee, denominated in the collateral
+/// token, into a decrease order's collateral withdrawal amount, so it can
+/// still be paid even when the position is closed at a loss. Returns the
+/// withdrawal amount including the estimate.
+///
+/// This is only an estimate used to size how much extra collateral to
+/// pull out of the position; the amount actually charged is computed
+/// afterwards, in the final output token, once it's known how much of
+/// that token the withdrawal (and any subsequent swap) actually produced.
+///
+/// Also rejects [`DecreasePositionSwapType::CollateralToPnlToken`]
+/// whenever a builder fee factor is set. This is different from ordinary
+/// underpayment: `NoSwap` can *sometimes* leave less in the final output
+/// token bucket than the fee needs (when the pnl token differs from the
+/// final output token), which is tolerated the same way any other
+/// decrease-side shortfall is. `CollateralToPnlToken`, however,
+/// deterministically moves the *entire* collateral-token output into the
+/// pnl token before the receive-token swap even runs (see
+/// `DecreasePosition::swap_collateral_token_to_pnl_token`), so the final
+/// output token bucket would be zero on every single order that chooses
+/// it. That's not variance, it's a guaranteed way to pay no fee at all,
+/// so it isn't allowed.
+///
+/// The rejection is keyed off the factor rather than off the estimate
+/// computed here. The estimate can round down to zero for a small
+/// requested size delta, while the fee actually charged is computed on
+/// the *executed* size delta, which a decrease may enlarge into a full
+/// close. Gating on a non-zero estimate would therefore admit exactly
+/// the orders that go on to incur a fee they can then bypass.
+fn estimate_builder_fee_for_collateral_withdrawal(
+    collateral_withdrawal_amount: u128,
+    size_delta_usd: u128,
+    builder_fee_factor: u128,
+    collateral_price: &Price<u128>,
+    decrease_position_swap_type: DecreasePositionSwapType,
+) -> Result<u128> {
+    if builder_fee_factor == 0 {
+        return Ok(collateral_withdrawal_amount);
+    }
+
+    require_neq!(
+        decrease_position_swap_type,
+        DecreasePositionSwapType::CollateralToPnlToken,
+        CoreError::BuilderFeeSwapTypeNotAllowed
+    );
+
+    let fee_estimate =
+        compute_builder_fee_amount(size_delta_usd, builder_fee_factor, collateral_price)?;
+
+    collateral_withdrawal_amount
+        .checked_add(fee_estimate)
+        .ok_or_else(|| error!(CoreError::TokenAmountOverflow))
+}
+
 #[allow(clippy::too_many_arguments)]
 #[inline(never)]
 fn execute_decrease_position(
@@ -2302,5 +2356,114 @@ mod tests {
         )
         .unwrap_err();
         assert_eq!(err, error!(CoreError::BuilderFeeExceedsCollateral));
+    }
+
+    #[test]
+    fn zero_builder_fee_factor_leaves_withdrawal_amount_unchanged() {
+        let withdrawal_amount = estimate_builder_fee_for_collateral_withdrawal(
+            1_000,
+            100_000 * constants::MARKET_USD_UNIT,
+            0,
+            &price(
+                2 * constants::MARKET_USD_UNIT,
+                2 * constants::MARKET_USD_UNIT,
+            ),
+            DecreasePositionSwapType::NoSwap,
+        )
+        .unwrap();
+        assert_eq!(withdrawal_amount, 1_000);
+    }
+
+    #[test]
+    fn tops_up_withdrawal_amount_by_builder_fee_estimate() {
+        // $100,000 size, 5 bps factor => $50 fee value, at $2/unit => 25 units.
+        let size_delta_usd = 100_000 * constants::MARKET_USD_UNIT;
+        let factor = constants::MARKET_USD_UNIT / 2_000;
+        let withdrawal_amount = estimate_builder_fee_for_collateral_withdrawal(
+            1_000,
+            size_delta_usd,
+            factor,
+            &price(
+                2 * constants::MARKET_USD_UNIT,
+                2 * constants::MARKET_USD_UNIT,
+            ),
+            DecreasePositionSwapType::PnlTokenToCollateralToken,
+        )
+        .unwrap();
+        // The withdrawal amount grows by the estimate; it is not subtracted.
+        assert_eq!(withdrawal_amount, 1_025);
+    }
+
+    #[test]
+    fn no_swap_is_allowed_with_a_non_zero_builder_fee() {
+        let size_delta_usd = 100_000 * constants::MARKET_USD_UNIT;
+        let factor = constants::MARKET_USD_UNIT / 2_000;
+        let result = estimate_builder_fee_for_collateral_withdrawal(
+            1_000,
+            size_delta_usd,
+            factor,
+            &price(
+                2 * constants::MARKET_USD_UNIT,
+                2 * constants::MARKET_USD_UNIT,
+            ),
+            DecreasePositionSwapType::NoSwap,
+        );
+        assert!(result.is_ok());
+    }
+
+    #[test]
+    fn collateral_to_pnl_token_is_rejected_with_a_non_zero_builder_fee() {
+        let size_delta_usd = 100_000 * constants::MARKET_USD_UNIT;
+        let factor = constants::MARKET_USD_UNIT / 2_000;
+        let result = estimate_builder_fee_for_collateral_withdrawal(
+            1_000,
+            size_delta_usd,
+            factor,
+            &price(
+                2 * constants::MARKET_USD_UNIT,
+                2 * constants::MARKET_USD_UNIT,
+            ),
+            DecreasePositionSwapType::CollateralToPnlToken,
+        );
+        assert!(result.is_err());
+    }
+
+    #[test]
+    fn collateral_to_pnl_token_is_rejected_when_the_fee_estimate_rounds_to_zero() {
+        // A dust size delta, whose fee value truncates to zero, while the
+        // factor is still set: the executed size delta may be enlarged
+        // into a full close and incur a real fee, so the swap type must
+        // already have been rejected by then.
+        let size_delta_usd = 1_000;
+        let factor = constants::MARKET_USD_UNIT / 2_000;
+        let result = estimate_builder_fee_for_collateral_withdrawal(
+            1_000,
+            size_delta_usd,
+            factor,
+            &price(
+                2 * constants::MARKET_USD_UNIT,
+                2 * constants::MARKET_USD_UNIT,
+            ),
+            DecreasePositionSwapType::CollateralToPnlToken,
+        );
+        assert_eq!(
+            result.unwrap_err(),
+            error!(CoreError::BuilderFeeSwapTypeNotAllowed)
+        );
+    }
+
+    #[test]
+    fn collateral_to_pnl_token_is_allowed_when_builder_fee_is_zero() {
+        let result = estimate_builder_fee_for_collateral_withdrawal(
+            1_000,
+            100_000 * constants::MARKET_USD_UNIT,
+            0,
+            &price(
+                2 * constants::MARKET_USD_UNIT,
+                2 * constants::MARKET_USD_UNIT,
+            ),
+            DecreasePositionSwapType::CollateralToPnlToken,
+        );
+        assert!(result.is_ok());
     }
 }
--- a/programs/store/src/ops/order.rs
+++ b/programs/store/src/ops/order.rs
@@ -4,7 +4,7 @@ use gmsol_callback::interface::ActionKind;
 use gmsol_model::{
     action::decrease_position::{DecreasePositionFlags, DecreasePositionSwapType},
     num::Unsigned,
-    price::Prices,
+    price::{Price, Prices},
     BaseMarket, BaseMarketExt, MarketAction, PnlFactorKind, Position as _, PositionMut,
     PositionMutExt, PositionState, PositionStateExt,
 };
@@ -12,6 +12,7 @@ use gmsol_utils::action::ActionCallbackKind;
 use typed_builder::TypedBuilder;
 
 use crate::{
+    constants,
     events::{EventEmitter, OrderUpdated, PositionDecreased, PositionIncreased, TradeData},
     states::{
         callback::CallbackAuthority,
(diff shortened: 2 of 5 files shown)


programs/store/src/ops/order.rs
```diff
@@ -2050,3 +2268,214 @@ impl PositionCutOperation<'_, '_> {
...
+        let factor = constants::MARKET_USD_UNIT / 2_000; // 5 bps
+        let unit_price = 2 * constants::MARKET_USD_UNIT;
+        let fee =
+            compute_builder_fee_amount(size_delta_usd, factor, &price(unit_price, unit_price))
+                .unwrap();
+        assert_eq!(fee, 25);
+    }
+
+    #[test]
+    fn rounds_up_on_remainder() {
```
angelo-75: nit: The test name does not reflect that this is a builder fee test.

crates/sdk/src/client/ops/exchange/order.rs
```diff
@@ -571,6 +571,15 @@ where
                         &short_token,
                         &token_program_id,
                         Some(&order),
+                    ))
+                    // Increase orders never swap out of the position, so
+                    // the final output token is always the collateral
+                    // token, but the escrow is still required so the
+                    // order can record it.
+                    .merge(self.client.prepare_associated_token_account(
```
angelo-75: issue: For increase orders, the final output token is currently always the collateral token, and the collateral token is currently always one of the long / short tokens, so this preparation is not necessary. It adds duplicate instruction calls and needlessly increases CU consumption. Please remove it.

programs/store/src/ops/order.rs
```diff
@@ -2050,3 +2268,214 @@ impl PositionCutOperation<'_, '_> {
...
+        let unit_price = 3 * constants::MARKET_USD_UNIT;
+        let fee =
+            compute_builder_fee_amount(size_delta_usd, factor, &price(unit_price, unit_price))
+                .unwrap();
+        // 50 / 3 = 16.67, rounded up.
+        assert_eq!(fee, 17);
+    }
+
+    #[test]
+    fn shortfall_clamps_to_available() {
```
angelo-75: nit: The test name does not reflect that this is a builder fee test.

programs/store/src/lib.rs
```diff
@@ -4251,6 +4251,27 @@ pub enum CoreError {
...
+    #[msg("builder fee requires the final output token to equal the collateral token")]
+    BuilderFeeFinalOutputTokenMismatch,
+    /// A decrease order with a non-zero builder fee used
+    /// `DecreasePositionSwapType::CollateralToPnlToken`. That swap type
+    /// deterministically moves the entire collateral-token output into the
+    /// pnl token before the receive-token swap runs, leaving the final
+    /// output token bucket at zero every time, which would make the
+    /// builder fee trivially and reliably bypassable.
+    #[msg("this decrease position swap type is not allowed together with a builder fee")]
+    BuilderFeeSwapTypeNotAllowed,
```
angelo-75: issue: These errors must be append-only. Please append at the end of the error list instead.

programs/store/src/ops/order.rs
```diff
@@ -2050,3 +2268,214 @@ impl PositionCutOperation<'_, '_> {
...
+        assert_eq!(clamp_builder_fee_amount(1_000, 100), 100);
+    }
+
+    #[test]
+    fn zero_available_clamps_to_zero() {
```
angelo-75: nit: The test name does not reflect that this is a builder fee test.

programs/store/src/ops/order.rs
```diff
@@ -1510,6 +1559,35 @@ fn execute_swap(
...
+/// cancelled), not a transaction revert.
+fn charge_builder_fee_on_collateral_increment(
+    collateral_increment_amount: u64,
+    size_delta_usd: u128,
+    builder_fee_factor: u128,
+    collateral_price: &Price<u128>,
+) -> Result<(u64, u128)> {
+    let payable_amount =
+        compute_builder_fee_amount(size_delta_usd, builder_fee_factor, collateral_price)?;
+    require!(
```
angelo-75: issue: Should use `require_gte!` instead.

programs/store/src/ops/order.rs
```diff
@@ -2050,3 +2268,214 @@ impl PositionCutOperation<'_, '_> {
...
+        assert_eq!(clamp_builder_fee_amount(1_000, 100), 100);
+    }
+
+    #[test]
+    fn zero_available_clamps_to_zero() {
+        assert_eq!(clamp_builder_fee_amount(1_000, 0), 0);
+    }
+
+    #[test]
+    fn fee_within_available_is_unchanged() {
```
angelo-75: nit: The test name does not reflect that this is a builder fee test.

programs/store/src/ops/order.rs
```diff
@@ -1510,6 +1559,35 @@ fn execute_swap(
...
+) -> Result<(u64, u128)> {
+    let payable_amount =
+        compute_builder_fee_amount(size_delta_usd, builder_fee_factor, collateral_price)?;
+    require!(
+        payable_amount <= u128::from(collateral_increment_amount),
+        CoreError::BuilderFeeExceedsCollateral
+    );
+    let collateral_increment_after_fee = collateral_increment_amount
+        .checked_sub(payable_amount as u64)
+        .ok_or_else(|| error!(CoreError::TokenAmountOverflow))?;
```
angelo-75, line 1591: nit: Consider using `CoreError::Internal` here, since the preceding assertion should guarantee that this can never fail.

programs/store/src/ops/order.rs
```diff
@@ -2050,3 +2268,214 @@ impl PositionCutOperation<'_, '_> {
...
+        assert_eq!(clamp_builder_fee_amount(1_000, 0), 0);
+    }
+
+    #[test]
+    fn fee_within_available_is_unchanged() {
+        assert_eq!(clamp_builder_fee_amount(10, 100), 10);
+    }
+
+    #[test]
+    fn zero_factor_leaves_collateral_increment_unchanged() {
```
angelo-75: nit: The test name does not reflect that this is a builder fee test.

programs/store/src/ops/order.rs
```diff
@@ -1510,6 +1559,35 @@ fn execute_swap(
...
+    collateral_price: &Price<u128>,
+) -> Result<(u64, u128)> {
+    let payable_amount =
+        compute_builder_fee_amount(size_delta_usd, builder_fee_factor, collateral_price)?;
+    require!(
+        payable_amount <= u128::from(collateral_increment_amount),
+        CoreError::BuilderFeeExceedsCollateral
+    );
+    let collateral_increment_after_fee = collateral_increment_amount
+        .checked_sub(payable_amount as u64)
```
angelo-75: issue: `u128 as u64` truncates silently. Our convention is to use `u64::try_from`.

programs/store/src/ops/order.rs
```diff
@@ -2050,3 +2268,214 @@ impl PositionCutOperation<'_, '_> {
...
+                2 * constants::MARKET_USD_UNIT,
+            ),
+        )
+        .unwrap();
+        assert_eq!(fee, 0);
+        assert_eq!(collateral_increment_after_fee, 1_000);
+    }
+
+    #[test]
+    fn charges_expected_fee() {
```
angelo-75: nit: The test name does not reflect that this is a builder fee test.

programs/store/src/ops/order.rs
```diff
@@ -1612,6 +1737,49 @@ fn execute_increase_position(
...
+    collateral_withdrawal_amount: u128,
+    size_delta_usd: u128,
+    builder_fee_factor: u128,
+    collateral_price: &Price<u128>,
+    decrease_position_swap_type: DecreasePositionSwapType,
+) -> Result<u128> {
+    let fee_estimate =
+        compute_builder_fee_amount(size_delta_usd, builder_fee_factor, collateral_price)?;
+
+    require!(
```
angelo-75: nit: Consider returning early when `fee_estimate == 0`, so that the require here becomes a single-condition check (use `require_neq!`).

programs/store/src/ops/order.rs
```diff
@@ -2050,3 +2268,214 @@ impl PositionCutOperation<'_, '_> {
...
+                2 * constants::MARKET_USD_UNIT,
+            ),
+        )
+        .unwrap();
+        assert_eq!(fee, 25);
+        assert_eq!(collateral_increment_after_fee, 975);
+    }
+
+    #[test]
+    fn exact_fee_is_allowed() {
```
angelo-75: nit: The test name does not reflect that this is a builder fee test.

programs/store/src/ops/order.rs
```diff
@@ -1624,14 +1792,26 @@ fn execute_decrease_position(
...
-        let collateral_withdrawal_amount = params.initial_collateral_delta_amount as u128;
         let size_delta_usd = params.size_delta_value;
         let acceptable_price = params.acceptable_price;
+        // Fold an estimate of the builder fee into the collateral
+        // withdrawal amount so it's still funded even if the position is
+        // closed at a loss. This is only an estimate for sizing the
+        // withdrawal; the amount actually charged is computed later, in
+        // the final output token, once the receive-token swap has run.
+        let collateral_withdrawal_amount = estimate_builder_fee_for_collateral_withdrawal(
+            params.initial_collateral_delta_amount as u128,
```
angelo-75: issue: Our convention here is to use `u128::from`.

programs/store/src/ops/order.rs
```diff
@@ -2050,3 +2268,214 @@ impl PositionCutOperation<'_, '_> {
...
+                2 * constants::MARKET_USD_UNIT,
+            ),
+        )
+        .unwrap();
+        assert_eq!(fee, 25);
+        assert_eq!(collateral_increment_after_fee, 0);
+    }
+
+    #[test]
+    fn underpayment_errors_instead_of_clamping() {
```
angelo-75: nit: The test name does not reflect that this is a builder fee test.

programs/store/src/ops/order.rs
```diff
@@ -1756,6 +1936,44 @@ fn execute_decrease_position(
...
+        // post-fee amount: the fee is simply netted out of what would
+        // otherwise be transferred out, so no separate accounting or
+        // balance-validation adjustment is needed. Underpayment is
+        // tolerated (charges whatever is available): unlike increase, this
+        // can't fall back to cancelling the order, since a decrease order
+        // may be closing an insolvent position.
+        let output_amount = if builder_fee_factor != 0 {
+            let final_output_token_price = oracle.get_primary_price(&final_output_token, false)?;
+            let payable_amount = compute_builder_fee_amount(
+                order.params.size_delta_value,
```
angelo-75: issue: The correct approach here is to use the `size_delta_usd` from the decrease report, because during the decrease it may be enlarged into a full close (in that case all of the collateral is effectively used for withdrawal, so this is the maximum withdrawable amount and nothing larger is possible, which means the slightly smaller estimate at the beginning does not affect the result here).

programs/store/src/ops/order.rs
```diff
@@ -2050,3 +2268,214 @@ impl PositionCutOperation<'_, '_> {
...
+                2 * constants::MARKET_USD_UNIT,
+                2 * constants::MARKET_USD_UNIT,
+            ),
+        )
+        .unwrap_err();
+        assert_eq!(err, error!(CoreError::BuilderFeeExceedsCollateral));
+    }
+
+    #[test]
+    fn zero_factor_leaves_withdrawal_amount_unchanged() {
```
angelo-75: nit: The test name does not reflect that this is a builder fee test.

programs/store/src/ops/order.rs
```diff
@@ -2050,3 +2268,214 @@ impl PositionCutOperation<'_, '_> {
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    fn price(min: u128, max: u128) -> Price<u128> {
+        Price { min, max }
+    }
+
+    #[test]
+    fn zero_factor_yields_zero_fee_without_needing_a_price() {
```
angelo-75: nit: The test name does not reflect that this is a builder fee test.

programs/store/src/ops/order.rs
```diff
@@ -2050,3 +2268,214 @@ impl PositionCutOperation<'_, '_> {
...
+                2 * constants::MARKET_USD_UNIT,
+            ),
+            DecreasePositionSwapType::NoSwap,
+        )
+        .unwrap();
+        assert_eq!(withdrawal_amount, 1_000);
+    }
+
+    #[test]
+    fn tops_up_withdrawal_amount_by_fee_estimate() {
```
angelo-75: nit: The test name does not reflect that this is a builder fee test.

programs/store/src/ops/order.rs
```diff
@@ -2050,3 +2268,214 @@ impl PositionCutOperation<'_, '_> {
...
+    }
+
+    #[test]
+    fn zero_factor_yields_zero_fee_without_needing_a_price() {
+        let fee = compute_builder_fee_amount(1_000_000, 0, &price(0, 0)).unwrap();
+        assert_eq!(fee, 0);
+    }
+
+    #[test]
+    fn computes_expected_amount() {
```
angelo-75: nit: The test name does not reflect that this is a builder fee test.

programs/store/src/ops/order.rs
```diff
@@ -2050,3 +2268,214 @@ impl PositionCutOperation<'_, '_> {
...
+            ),
+            DecreasePositionSwapType::PnlTokenToCollateralToken,
+        )
+        .unwrap();
+        // The withdrawal amount grows by the estimate; it is not subtracted.
+        assert_eq!(withdrawal_amount, 1_025);
+    }
+
+    #[test]
+    fn no_swap_is_allowed_with_a_non_zero_fee() {
```
angelo-75: nit: The test name does not reflect that this is a builder fee test.

programs/store/src/ops/order.rs
```diff
@@ -2050,3 +2268,214 @@ impl PositionCutOperation<'_, '_> {
...
+                2 * constants::MARKET_USD_UNIT,
+                2 * constants::MARKET_USD_UNIT,
+            ),
+            DecreasePositionSwapType::NoSwap,
+        );
+        assert!(result.is_ok());
+    }
+
+    #[test]
+    fn collateral_to_pnl_token_is_rejected_with_a_non_zero_fee() {
```
angelo-75: nit: The test name does not reflect that this is a builder fee test.

programs/store/src/ops/order.rs
```diff
@@ -2050,3 +2268,214 @@ impl PositionCutOperation<'_, '_> {
...
+                2 * constants::MARKET_USD_UNIT,
+                2 * constants::MARKET_USD_UNIT,
+            ),
+            DecreasePositionSwapType::CollateralToPnlToken,
+        );
+        assert!(result.is_err());
+    }
+
+    #[test]
+    fn collateral_to_pnl_token_is_allowed_when_fee_is_zero() {
```
angelo-75: nit: The test name does not reflect that this is a builder fee test.

angelo-75 (changes requested):

programs/store/src/ops/order.rs
gmitsj: Renamed to builder_fee_amount_rounds_up_on_remainder

programs/store/src/lib.rs
gmitsj: Good catch, fixed. All three now live in a new Builder Fee Errors section at the very end of `CoreError`, right above the append-only NOTE, so no existing code shifts. Each of the three commits appends its own variant there rather than inserting into the order section. I also corrected `crates/programs/idls/gmsol_store.json`, which had picked up the shifted codes.

programs/store/src/ops/order.rs
gmitsj: Agreed, changed to `report.size_delta_usd()`. I've recorded your reasoning in a comment at the call site. The executed size delta can be larger than the requested one when the decrease is enlarged into a full close, but in that case the entire collateral is withdrawn, so the fee computed from the executed size is still within what the withdrawal can support, the smaller estimate used earlier to size the withdrawal is not a binding constraint. The estimate call itself still uses `params.size_delta_value`, since it runs before `decrease()`

programs/store/src/lib.rs
```diff
@@ -4407,6 +4407,27 @@ pub enum CoreError {
...
+    /// builder fee. Unlike decrease, increase has no equivalent of
+    /// "complete anyway at a loss" to fall back on, so this cancels the
+    /// order instead of charging a partial fee.
+    #[msg("builder fee exceeds the collateral increment")]
+    BuilderFeeExceedsCollateral,
+    /// A non-zero builder fee is being charged on an increase order whose
+    /// final output token is not the collateral token. This should be
+    /// unreachable for orders created after final output token
+    /// initialization was added for increase orders, and guards against
+    /// orders created before that.
```
angelo-75: issue: The second sentence is not entirely accurate, so I suggest removing or rewriting it. The fact is that the initialization of the final output token is still optional.

programs/store/src/ops/order.rs
```diff
@@ -1510,6 +1559,39 @@ fn execute_swap(
...
+        u128::from(collateral_increment_amount),
+        payable_amount,
+        CoreError::BuilderFeeExceedsCollateral
+    );
+    // The assertion above guarantees both the conversion and the
+    // subtraction below succeed.
+    let collateral_increment_after_fee = u64::try_from(payable_amount)
+        .ok()
+        .and_then(|fee| collateral_increment_amount.checked_sub(fee))
+        .ok_or_else(|| error!(CoreError::Internal))?;
```
angelo-75: issue: A failed `u64::try_from` here should produce an overflow error rather than an internal error. Besides, since the u64 version of payable_amount is used below anyway, I suggest:

```rust
let payable_amount_u64 = u64::try_from(payable_amount).map_err(|_| error!(CoreError::TokenAmountOverflow))?;

require_gte!(collateral_increment_amount, payable_amount_u64, CoreError::BuilderFeeExceedsCollateral);

// ...
```

programs/store/src/ops/order.rs
```diff
@@ -1567,6 +1650,56 @@ fn execute_increase_position(
...
+                builder_fee_factor,
+                position.collateral_price(&prices),
+            )?;
+
+        // TODO(builder-fee): accumulate `payable_amount` into
+        // `order.builder_fee_amount` once that field exists.
+        // The helper only ever returns a fee it could subtract from the
+        // (`u64`) collateral increment, so this conversion cannot fail.
+        let payable_amount_u64 =
+            u64::try_from(payable_amount).map_err(|_| error!(CoreError::Internal))?;
```
angelo-75: issue: use `CoreError::TokenAmountOverflow` instead.

angelo-75 (changes requested):
There are still four issue-level comments left unaddressed.

crates/sdk/src/client/ops/exchange/order.rs
gmitsj: Removed. You're right, and the branch was contradicting itself. The comment directly below the code you flagged already says opting into the final output token escrow needs no extra preparation, because for an increase order that token is the collateral token and therefore one of the long/short escrows prepared just above.

programs/store/src/lib.rs
gmitsj: Fixed, and the same wrong claim was in the caller comment in execute_increase_position, so both are rewritten.

programs/store/src/ops/order.rs
gmitsj: The conversion is gone rather than re-tagged. Your point that the `u64` version is used below anyway holds across the helper boundary too, so `charge_builder_fee_on_collateral_increment` now returns `(u64, u64)` and the caller has nothing left to convert.

programs/store/src/ops/order.rs
gmitsj: Adopted your snippet as written, `payable_amount_u64` binding included

crypto1gmt:
issue: the builder fee is netted in the decrease path but routed in the increase path, and the two stop agreeing once #408 lands.

increase (1682) puts the fee in the transfer bucket, so it physically moves market vault -> order escrow:

```rust
transfer_out.transfer_out(false, payable_amount)?;
```

decrease (1974, 2001) subtracts it from the payout and routes only the remainder:

```rust
let output_amount = ... output_amount.checked_sub(paid)?;
...
transfer_out.transfer_out(false, output_amount)?;
```

there's no `transfer_out(false, paid_amount)` anywhere in the function, so the fee value never leaves the market vault.

that's self-consistent on its own, and the comment at 1958 says as much. the problem is that both TODOs (1680, 1979) accumulate into `order.builder_fee_amount`, and #408's `settle_builder_fee` pays the builder out of the escrow:

```rust
transfer_checked(... from: escrow, to: claim_vault ..., settled_amount)
```

so once both land, increase balances (escrow +fee, settlement -fee) but decrease doesn't: the escrow is already short by the fee, settlement takes it again out of the reduced escrow, the user pays twice and the withheld amount sits in the market vault.

worth flagging that the clamp in `settle_builder_fee` doesn't catch this. `recorded_amount.min(escrow.amount)` only bites when the escrow is smaller than the fee, and here the escrow holds `output_amount - fee`, which normally still covers it. so it fails silently rather than erroring out. the invariant that comment states, "the escrow always covers the recorded amount", is exactly what the decrease path breaks.

fix direction is presumably to stop netting and let settlement be the only thing that pays the builder: route the full `output_amount` into the bucket, record `paid_amount`, and let settlement split it. that does move what `validate_decrease_output_amounts` sees, so it's worth deciding together with #408 rather than in isolation.

credit to @mv-reyes, who spotted this on #416.

programs/store/src/ops/order.rs
```diff
@@ -1510,6 +1559,39 @@ fn execute_swap(
...
+/// cancelled), not a transaction revert.
+fn charge_builder_fee_on_collateral_increment(
+    collateral_increment_amount: u64,
+    size_delta_usd: u128,
+    builder_fee_factor: u128,
+    collateral_price: &Price<u128>,
+) -> Result<(u64, u64)> {
+    let payable_amount =
+        compute_builder_fee_amount(size_delta_usd, builder_fee_factor, collateral_price)?;
+    let payable_amount_u64 =
```
angelo-75: nit: shadowing may be enough this time

programs/store/src/ops/order.rs
```diff
@@ -1567,6 +1650,52 @@ fn execute_increase_position(
...
+    // `validate_market_balances` further down does not need to be
+    // adjusted to cover it.
+    let collateral_increment_amount = if builder_fee_factor != 0 {
+        // Presence only: the check above already guarantees equality with
+        // the collateral token whenever `final_output_token` is `Some`.
+        // Initializing it is optional for increase orders (see
+        // `CreateIncreaseOrderOperation`), so an order whose creator did
+        // not opt in carries `None` here and cannot pay a fee.
+        require!(
+            order.tokens.final_output_token.token().is_some(),
```
angelo-75: suggestion: It would be better to validate it more specifically as Some(collateral_token) here.

programs/store/src/ops/order.rs
```diff
@@ -1756,6 +1944,53 @@ fn execute_decrease_position(
...
+                &position.market().market_meta().market_token_mint,
+                &final_output_token,
+                payable_amount,
+                paid_amount,
+            )?)?;
+
+            output_amount
+        } else {
+            output_amount
+        };
```
angelo-75: issue: As @crypto1gmt pointed out, paid_amount does not need to be deducted from output_amount here; it only needs to be recorded into buyback_fee_amount in the future. So this code can be slightly refactored: no modification to output_amount is needed at all. Instead, just introduce a separate builder_fee_factor branch that computes payable_amount and paid_amount, records them into builder_fee_amount, and emits the event.

angelo-75 (changes requested):

programs/store/src/ops/order.rs
gmitsj: shadowed into `payable_amount`

programs/store/src/ops/order.rs
gmitsj: Changed to `require!(order.tokens.final_output_token.token() == Some(*position.collateral_token()), ...)`. The check at the top of the function already rejects a `Some` that disagrees with the collateral token, so this is the same condition stated positively, but it makes the fee branch depend on what it actually needs rather than on a non-local argument, and the comment was updated to match.

gmitsj:
@crypto1gmt @angelo-75 
The decrease path no longer nets, the full `output_amount` goes into the final-output-token bucket, so the fee value physically lands in the escrow exactly as it does on the increase side, and `settle_builder_fee` is the only thing that pays the builder. `paid_amount` is still clamped to `output_amount`, which is now precisely what the escrow will hold, so the "escrow always covers the recorded amount" invariant is true by construction on both paths rather than by assumption.

On the point about `validate_decrease_output_amounts`: it now sees the gross amount, i.e. `min_output` is checked against the payout before the builder fee rather than after. That falls out of routing instead of netting. It reads as the right semantics to me, `min_output` is the user's slippage bound on the trade, and the builder fee is a separate charge they opted into, but flagging it explicitly since you asked for it to be decided alongside #408 rather than silently.

The collateral-withdrawal top-up in `estimate_builder_fee_for_collateral_withdrawal` is unchanged and still does its job. The withdrawal is sized to include the fee, so the escrow is funded even when the position closes at a loss.

programs/store/src/ops/order.rs
gmitsj: `output_amount` is untouched, and the `builder_fee_factor` branch now only computes payable_amount/paid_amount, emits the event, and carries the TODO for recording into `builder_fee_amount`. The branch no longer produces a value, so the else arm is gone too.

mv-reyes:
went ahead and wrote the fix for the decrease-path routing plus a regression test, since I had the analysis from the read-through on #416.

the fix: don't net the fee before transfer_out. transfer the full pre-fee output to the escrow, accumulate the recorded fee, and keep slippage validation on the post-fee amount. same pattern the increase path already uses at order.rs:1686. with that, settle pays the builder from the escrow and close sweeps exactly output - fee to the user, nothing stranded.

repro: https://github.com/mv-reyes/gmsol-builder-fee-repro (cargo test). current_wiring_charges_user_twice documents the behavior as written: user ends up with output - 2*fee and the fee amount is stranded in the market vault. fixed_routing_charges_once passes with the patch applied. deep_loss_close_does_not_overdraw covers the clamped edge.

happy to open the actual patch as a pr against this branch if thats easier to review.

mv-reyes:
just saw the routing fix landed above, that matches what I had. nice.

my repro still works as the regression pin for it if you want one: cargo test in https://github.com/mv-reyes/gmsol-builder-fee-repro shows the pre-fix behavior (user charged twice, fee stranded in the vault) and the fixed routing passing. its a self-contained model of the flow, not wired to your harness. if you want it ported into the anchor test suite as a proper regression test I can pr that.

programs/store/src/ops/order.rs
angelo-75: nit: `require_eq!` might be a better fit here? Not blocking the merge on this one this time though.

angelo-75:
> On the point about `validate_decrease_output_amounts`: it now sees the gross amount, i.e. `min_output` is checked against the payout before the builder fee rather than after. That falls out of routing instead of netting. It reads as the right semantics to me, `min_output` is the user's slippage bound on the trade, and the builder fee is a separate charge they opted into, but flagging it explicitly since you asked for it to be decided alongside #408 rather than silently.

suggestion: I think your reasoning is sound, and I also agree that keeping the current behavior (with the builder fee included) is semantically more complete. But it is worth leaving a comment noting that when a builder implements slippage control via `min_output`, they need to take the builder fee into account as well (it will be part of the output). Not blocking this merge on it.

angelo-75:
@mv-reyes 

Thank you for the contribution, we really appreciate it. That said, as a general rule, we don't directly merge code from outside contributors.

programs/store/src/ops/order.rs
```diff
@@ -1612,6 +1743,53 @@ fn execute_increase_position(
...
+
+    if fee_estimate == 0 {
+        return Ok(collateral_withdrawal_amount);
+    }
+
+    require_neq!(
+        decrease_position_swap_type,
+        DecreasePositionSwapType::CollateralToPnlToken,
+        CoreError::BuilderFeeSwapTypeNotAllowed
+    );
```
angelo-75, line 1801: issue: This check should be unconditional (i.e., it should come before the early-return branch). Otherwise, if fee_estimate is indeed computed as 0 because size_delta_usd is too small, and the position later gets converted into a full close, a builder fee may still end up being incurred; but by then, since the CollateralToPnlToken swap has been allowed, the builder fee could potentially be bypassed.

programs/store/src/ops/order.rs
gmitsj: Went with `require_keys_eq!` rather than `require_eq!` because the values are `Pubkeys`, and `require_eq!` on `Option<Pubkey>` doesn't compile anyway since `with_values` needs `Display` on both sides. So the `Option` is unwrapped first into the same error, and the check reads the same way as the equality check at the top of the function.

programs/store/src/ops/order.rs
gmitsj: The check now runs before anything else and is keyed off builder_fee_factor rather than off the estimate; only the top-up itself is skipped when there's no fee

angelo-75 (approved):

PR #407, opened 2026-08-12
feat: add builder fee state and config carve-outs

gmitsj:
## Summary

Carves the builder fee state fields into `Order`, `UserHeader`, and the store config, adds the builder fee feature flag and factor cap config keys, and defines the `UserTokenController` PDA derivation, per ADR-0001. Everything here is passive. No instruction reads or writes any of these fields yet, so this can merge ahead of and in parallel with the settlement, configuration, and activation tracks.

Every field addition is carved out of an existing `reserved` region. Account sizes and every pre-existing field's offset are unchanged, verified with `core::mem::offset_of!` assertions rather than just claimed, so live accounts need no migration.

## What's implemented

- `Order` gains `builder: Pubkey`, `builder_fee_factor: u128`, and `builder_fee_amount: u64`, carved from `reserved: [u8; 128]` down to `[u8; 72]`. Field order was chosen so the `u128` lands on a 16-byte-aligned offset with zero compiler-inserted padding, matched exactly by a test that checks the real offsets. Read-only getters `builder()`, `builder_fee_factor()`, `builder_fee_amount()`. Nothing writes them yet.
- `UserHeader` gains `builder_fee_factor: u128`, carved from `reserved: [u8; 128]` down to `[u8; 112]`, with a read-only getter. This is the builder's advertised rate, distinct from the snapshot copy on `Order`.
- `Store::Factors` gains `max_builder_fee_factor: Factor`, carved from `reserved: [Factor; 64]` down to `[Factor; 63]`, wired into the existing `get`/`get_mut` match arms. No `init()` default, so it reads as `0` on every existing store. `insert_factor` already dispatches generically by string key, so no instruction-level change was needed to make it settable.
- `FactorKey::MaxBuilderFeeFactor` and `DomainDisabledFlag::BuilderFee` appended in `crates/utils/src/config.rs`. `DisabledFeatures` turned out to already be a generic `(domain, action)` map rather than a hardcoded match, so the flag needed no extra wiring beyond the enum variant itself.
- `UserTokenController` PDA derivation, `[SEED, user_account, token_mint]`, defined as a program-side seed constant (`programs/store/src/states/user.rs`) and an SDK helper (`find_user_token_controller_address` in `crates/sdk/src/pda.rs`). No account struct, no init instruction. The per-user component is the User Account PDA rather than a raw wallet address, since it's already store-scoped and matches the identity convention the rest of ADR-0001 uses.

## Testing

- `cargo test -p gmsol-store`. 16 passed, 0 failed, up from 9 pre-existing.
- For each touched struct (`Order`, `UserHeader`, `Factors`), a layout test pair. One asserts pre-existing fields keep their exact offsets and the struct's total size is unchanged. One asserts the new fields sit inside what used to be the `reserved` region, at correctly aligned offsets.
- A separate test per struct constructs a zeroed instance via `bytemuck::Zeroable::zeroed()`, standing in for a real pre-upgrade on-chain account (Solana account data is zero-initialized, and nothing ever wrote into these `reserved` regions), and asserts the new getters read back as the "no builder fee" defaults. This exercises the acceptance criteria directly rather than only arguing it from the offset math.
- `cargo build` / `cargo check --workspace` / `cargo clippy -p gmsol-store -p gmsol-sdk -p gmsol-utils --lib`. Clean, no new warnings.

## Out of scope

- Any instruction that reads or writes these fields. That's the settlement, configuration, and activation tracks this PR is meant to unblock.
- The `UserTokenController` account struct and its init instruction. Per review, this issue only fixes the PDA derivation.
- IDL regeneration. The new `#[constant]` won't appear in `crates/programs/idls/gmsol_store.json` until an `anchor build` picks it up. Nothing in this PR depends on that having happened.

### DIFF
--- a/crates/sdk/src/pda.rs
+++ b/crates/sdk/src/pda.rs
@@ -17,7 +17,8 @@ use gmsol_programs::gmsol_timelock::accounts as timelock_accounts;
 use gmsol_programs::gmsol_liquidity_provider::accounts as liquidity_provider_accounts;
 
 pub use gmsol_programs::gmsol_store::constants::{
-    VIRTUAL_INVENTORY_FOR_POSITIONS_SEED, VIRTUAL_INVENTORY_FOR_SWAPS_SEED,
+    USER_TOKEN_CONTROLLER_SEED, VIRTUAL_INVENTORY_FOR_POSITIONS_SEED,
+    VIRTUAL_INVENTORY_FOR_SWAPS_SEED,
 };
 
 /// Nonce bytes.
@@ -73,11 +74,6 @@ pub const USER_SEED: &[u8] = b"user";
 
 /// Seed for [`ReferralCodeV2`](store_accounts::ReferralCodeV2).
 pub const REFERRAL_CODE_SEED: &[u8] = b"referral_code";
-
-/// Seed for the (per-user, per-token) user token controller PDA.
-///
-/// This PDA has no backing account yet; only its derivation is fixed.
-pub const USER_TOKEN_CONTROLLER_SEED: &[u8] = b"user_token_controller";
 
 /// Seed for GLV token mint.
 pub const GLV_TOKEN_SEED: &[u8] = b"glv_token";
--- a/crates/sdk/src/pda.rs
+++ b/crates/sdk/src/pda.rs
@@ -74,6 +74,11 @@ pub const USER_SEED: &[u8] = b"user";
 /// Seed for [`ReferralCodeV2`](store_accounts::ReferralCodeV2).
 pub const REFERRAL_CODE_SEED: &[u8] = b"referral_code";
 
+/// Seed for the (per-user, per-token) user token controller PDA.
+///
+/// This PDA has no backing account yet; only its derivation is fixed.
+pub const USER_TOKEN_CONTROLLER_SEED: &[u8] = b"user_token_controller";
+
 /// Seed for GLV token mint.
 pub const GLV_TOKEN_SEED: &[u8] = b"glv_token";
 
@@ -342,6 +347,31 @@ pub fn find_referral_code_address(
 ) -> (Pubkey, u8) {
     Pubkey::find_program_address(
         &[REFERRAL_CODE_SEED, store.as_ref(), &code],
+        store_program_id,
+    )
+}
+
+/// Find PDA for the (per-user, per-token) user token controller.
+///
+/// `user_account` is the [`User`](store_accounts::UserHeader) account's
+/// own PDA (see [`find_user_address`]), not the user's wallet address; it
+/// is already store-scoped, so no separate store seed is needed here. This
+/// mirrors the derivation keys of the claim ATA the controller will guard
+/// (`(user_account, token_mint)`), and matches the identity convention
+/// where a builder is identified by its User Account.
+///
+/// This PDA has no backing account yet; only its derivation is fixed.
+pub fn find_user_token_controller_address(
+    user_account: &Pubkey,
+    token_mint: &Pubkey,
+    store_program_id: &Pubkey,
+) -> (Pubkey, u8) {
+    Pubkey::find_program_address(
+        &[
+            USER_TOKEN_CONTROLLER_SEED,
+            user_account.as_ref(),
+            token_mint.as_ref(),
+        ],
         store_program_id,
     )
 }
--- a/programs/store/src/states/order.rs
+++ b/programs/store/src/states/order.rs
@@ -325,9 +325,16 @@ pub struct Order {
     pub(crate) gt_reward: u64,
     #[cfg_attr(feature = "debug", debug(skip))]
     padding_1: [u8; 8],
+    /// The builder attached to this order, at order creation time. The
+    /// default (zero) [`Pubkey`] means no builder is attached.
+    pub(crate) builder: Pubkey,
+    /// The builder's fee factor, snapshotted when the builder fee is set.
+    pub(crate) builder_fee_factor: u128,
+    /// The builder fee amount charged so far, pending settlement.
+    pub(crate) builder_fee_amount: u64,
     #[cfg_attr(feature = "debug", debug(skip))]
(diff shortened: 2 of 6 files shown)


programs/store/src/states/order.rs
```diff
@@ -325,9 +325,16 @@ pub struct Order {
     pub(crate) gt_reward: u64,
     #[cfg_attr(feature = "debug", debug(skip))]
     padding_1: [u8; 8],
```
angelo-75: nit: actually, we could consider replacing this field with `builder_fee_amount`

programs/store/src/states/order.rs
```diff
@@ -943,3 +965,52 @@ impl OrderActionParams {
             .set_flag(OrderFlag::ShouldKeepPositionAccount, keep)
     }
 }
+
+#[cfg(test)]
+mod builder_fee_layout_tests {
```
angelo-75: nit: as mentioned before, we don't use per-section tests mods. Instead, use a single tests mod placed at the end of the `.rs` file.

programs/store/src/states/order.rs
```diff
@@ -943,3 +965,52 @@ impl OrderActionParams {
...
+    use super::*;
+
+    // Captured from the layout before the builder fee fields were carved
+    // out of `reserved`. These must never change.
+    const OLD_SIZE: usize = 2464;
```
angelo-75: nit: since it never changes, the naming should reflect that, e.g. EXPECTED_ORDER_ACCOUNT_SIZE

programs/store/src/states/order.rs
```diff
@@ -943,3 +965,52 @@ impl OrderActionParams {
...
+    const GT_REWARD_OFFSET: usize = 2320;
+    const PADDING_1_OFFSET: usize = 2328;
+    const OLD_RESERVED_OFFSET: usize = 2336;
```
angelo-75: nit: from a long-term maintenance perspective, this naming is not ideal. Consider BUILDER_OFFSET instead.

programs/store/src/states/order.rs
```diff
@@ -943,3 +965,52 @@ impl OrderActionParams {
...
+
+    #[test]
+    fn existing_fields_keep_their_offsets_and_size_is_unchanged() {
+        assert_eq!(std::mem::size_of::<Order>(), OLD_SIZE);
+        assert_eq!(std::mem::offset_of!(Order, gt_reward), GT_REWARD_OFFSET);
+        assert_eq!(std::mem::offset_of!(Order, padding_1), PADDING_1_OFFSET);
+    }
+
+    #[test]
+    fn builder_fee_fields_are_carved_from_the_old_reserved_region_with_correct_alignment() {
```
angelo-75: nit: actually, this kind of test has little value in the long run. Tests should verify the expected state rather than the change itself. Consider removing this test, or replacing it with an assertion on the expected format.

programs/store/src/states/store.rs
```diff
@@ -647,12 +652,62 @@ impl Factors {
...
             }
+            FactorKey::MaxBuilderFeeFactor => &mut self.max_builder_fee_factor,
             _ => return None,
         };
         Some(value)
     }
 }
 
+#[cfg(test)]
+mod factors_layout_tests {
```
angelo-75: nit: please make the changes following the review comments on builder_fee_layout_tests

programs/store/src/states/user.rs
```diff
@@ -340,3 +359,42 @@ impl UserGtState {
         self.amount
     }
 }
+
+#[cfg(test)]
+mod builder_fee_layout_tests {
```
angelo-75: nit: please make the changes following the review comments on builder_fee_layout_tests

crates/sdk/src/pda.rs
```diff
@@ -74,6 +74,11 @@ pub const USER_SEED: &[u8] = b"user";
 /// Seed for [`ReferralCodeV2`](store_accounts::ReferralCodeV2).
 pub const REFERRAL_CODE_SEED: &[u8] = b"referral_code";
 
+/// Seed for the (per-user, per-token) user token controller PDA.
+///
+/// This PDA has no backing account yet; only its derivation is fixed.
+pub const USER_TOKEN_CONTROLLER_SEED: &[u8] = b"user_token_controller";
```
angelo-75: suggestion: since USER_TOKEN_CONTROLLER_SEED is already exported as a constant, it should be loaded from the programs crate rather than redefined.

angelo-75 (dismissed):

programs/store/src/states/order.rs
```diff
@@ -323,11 +323,16 @@ pub struct Order {
...
     pub(crate) params: OrderActionParams,
     pub(crate) gt_reward: u64,
-    #[cfg_attr(feature = "debug", debug(skip))]
-    padding_1: [u8; 8],
+    /// The builder fee amount charged so far, pending settlement.
+    pub(crate) builder_fee_amount: u64,
+    /// The builder attached to this order, at order creation time. The
+    /// default (zero) [`Pubkey`] means no builder is attached.
+    pub(crate) builder: Pubkey,
+    /// The builder's fee factor, snapshotted at order creation time.
```
angelo-75: issue (blocking): it is not snapshotted at order creation time; it is snapshotted when the builder fee is set.

programs/store/src/states/order.rs
```diff
@@ -534,6 +539,21 @@ impl Order {
         &self.tokens
     }
 
+    /// Get the builder attached to this order, if any.
+    pub fn builder(&self) -> Option<&Pubkey> {
+        optional_address(&self.builder)
+    }
+
+    /// Get the builder's fee factor, snapshotted at order creation time.
```
angelo-75: issue (blocking): it is not snapshotted at order creation time; it is snapshotted when the builder fee is set.

angelo-75 (changes requested):
All the issues found in the last review have been resolved. However, this review noticed incorrect comments (these would get flagged by auditors), so they need to be fixed.

angelo-75 (approved):

PR #408, opened 2026-08-13
feat: builder fee settlement rails, close guard and permissionless settle_builder_fee

gmitsj:
CON-37: Add builder fee settlement rails: close guard and permissionless settle_builder_fee

## Summary

Adds the two rails ADR-0001 needs before builder fee charging can be safely activated. Once an order executes with a builder fee, the charged amount sits in the order's final output token escrow and is recorded in `order.builder_fee_amount`. Closing an order wipes its on-chain data, so closing one with an unsettled fee would sweep it to the receiver along with the regular output and destroy the beneficiary information. This PR prevents that with a close guard and a dedicated, permissionless `settle_builder_fee` instruction.

Everything here is safe to merge immediately, ahead of activation. Accrual is still inert (`builder_fee_amount` is always 0 in production), so the guard never triggers and settlement always no-ops until a later PR turns real charging on.

Depends on `feat/builder-fee-state-v2` (`Order.builder` / `Order.builder_fee_amount`), hence this PR targets that branch rather than `main`.

## What's implemented

- **Close guard**, in `CloseOrderV2::validate()` (`programs/store/src/instructions/exchange/order.rs`), which Anchor calls before both close paths (owner close in any state, keeper close of terminal orders per `internal::Close::close()`). Rejects with the new `CoreError::UnsettledBuilderFee` whenever `order.builder_fee_amount() != 0`, appended at the end of the `CoreError` enum per the existing "new variants must be appended" convention, so it doesn't shift any existing error codes.
- **`BuilderFeeSettled` event** (`programs/store/src/events/builder_fee.rs`), recording the order, the builder, the token mint, and both the recorded and the actually-settled amounts.
- **`settle_builder_fee` instruction** (`programs/store/src/instructions/builder_fee.rs`), permissionless (no signer role, no `#[access_control]`), transferring `min(builder_fee_amount, escrow.amount)` from the order's final output token escrow to the builder's claim vault (the associated token account of the final output token owned by the builder's User Account), then unconditionally resetting `builder_fee_amount` to 0. `builder_user` and `claim_vault` are both `Option`, required only when the recorded amount is non-zero. When it's 0, including orders that never had a builder set, the handler returns immediately: no deserialization of those accounts, no CPI, no event.
- **SDK support** (`crates/sdk/src/client/ops/builder_fee.rs`, `BuilderFeeOps::settle_builder_fee`), which fetches the order when no hint is given and correctly derives `None`/`None` for the builder accounts on the no-op path rather than needing a placeholder account.

## Testing

- `cargo build --workspace` / `cargo test -p gmsol-store -p gmsol-sdk --lib` / `cargo clippy -p gmsol-store -p gmsol-sdk --lib`. All clean, 34 + 13 tests passing, no new warnings.
- No colocated Rust unit tests added: there's no pure logic here worth extracting (the clamp is a one-line `.min()`), consistent with how `CreateOrderV2`/`CloseOrderV2` and the rest of this instruction layer are already tested only through the `anchor_test` integration suite, not Rust unit tests.
- Did not run the full `anchor_test` suite for this PR (no test scenario exists yet for an order with a non-zero builder fee, since activation hasn't landed), so the close-guard and settlement paths are exercised by code review and by inspection against the acceptance criteria here rather than an integration test run.

### DIFF
--- a/crates/programs/idls/gmsol_store.json
+++ b/crates/programs/idls/gmsol_store.json
@@ -7421,7 +7421,11 @@
           "name": "final_output_token_escrow",
           "docs": [
             "Final output token escrow account.",
-            "Only required by decrease and swap orders."
+            "",
+            "Required by decrease and swap orders. Optional for increase orders: when provided, it is",
+            "recorded on the order and becomes the account a builder fee would be paid out of, and when",
+            "omitted the order's final output token is left uninitialized, so no builder fee can be set",
+            "on it later."
           ],
           "writable": true,
           "optional": true,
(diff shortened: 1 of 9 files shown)


programs/store/src/instructions/builder_fee.rs
```diff
@@ -0,0 +1,136 @@
...
+    /// need it, so it may be omitted.
+    pub builder_user: Option<AccountLoader<'info, UserHeader>>,
+    /// The builder's claim vault: the associated token account of the
+    /// final output token owned by the builder's User Account.
+    ///
+    /// Required to already exist (its existence is validated when the
+    /// builder sets their fee), so no `init_if_needed` is used here.
+    /// Only required alongside `builder_user`, for the same reason.
+    #[account(mut)]
+    pub claim_vault: Option<Box<Account<'info, TokenAccount>>>,
```
crypto1gmt, line 60: issue (non-blocking): the vault is pinned by owner and mint, but not to the associated token account the comment above describes.

Owner plus mint does not identify a single account: a token account can be created with any `owner` field, so with `settle_builder_fee` being permissionless the caller gets to choose which of them receives the fee, and the claim side would be looking at the ATA. ADR-0001 says settlement moves the fee "into the builder's User Account ATA", #407's `find_user_token_controller_address` doc describes the controller as mirroring "the derivation keys of the claim ATA", and #389 had `associated_token::mint` / `associated_token::authority` on this account.

Not blocking, since `builder_fee_amount` is always 0 until CON-41, so there is nothing to misroute yet. It would want to be true before charging is switched on, though. The `escrow` account four lines up already carries the constraint, so mirroring it here and dropping the two `require_keys_eq!` in the body would line the two up.

programs/store/src/instructions/builder_fee.rs
```diff
@@ -0,0 +1,136 @@
...
+    ///
+    /// Only required when the order has a non-zero recorded builder fee
+    /// amount. The no-op path (a recorded amount of zero, including
+    /// orders that never had a builder set) performs no CPI and does not
+    /// need it, so it may be omitted.
+    pub builder_user: Option<AccountLoader<'info, UserHeader>>,
+    /// The builder's claim vault: the associated token account of the
+    /// final output token owned by the builder's User Account.
+    ///
+    /// Required to already exist (its existence is validated when the
```
crypto1gmt: issue (non-blocking): "its existence is validated when the builder sets their fee" does not hold yet.

I went through CON-40's ACs this morning, since it is the next one on my side, and `set_builder_fee` is not written; the name only appears inside comments about future work. CON-39's `set_builder_fee_factor` checks the factor cap and `is_initialized()` and touches no token account.

Worth fixing while it is cheap, for two reasons. Comments like this tend to get picked up in audits. And it is load-bearing for the close guard added in this PR: if the vault does not exist, the `Option<Box<Account<TokenAccount>>>` cannot deserialize, settlement fails, and `CloseOrderV2::validate()` then rejects the close, with `close_order_v2` being the only close path. Inert today for the same reason as above, but it becomes real the moment a fee can be non-zero.

`init_if_needed` with a payer would close it (the feature is already on `anchor-lang` and used in `exchange/deposit.rs` and `exchange/execute_order.rs`), or CON-40 could carry the guarantee instead. I am happy to make CON-40 carry it, in which case this comment could point at it as future work rather than as something that already happens. What do you think?

programs/store/src/instructions/builder_fee.rs
```diff
@@ -0,0 +1,136 @@
...
+        constraint = order.load()?.tokens().final_output_token.account() == Some(escrow.key()) @ CoreError::TokenAccountMismatched,
+    )]
+    pub escrow: Box<Account<'info, TokenAccount>>,
+    /// The builder's User Account.
+    ///
+    /// Only required when the order has a non-zero recorded builder fee
+    /// amount. The no-op path (a recorded amount of zero, including
+    /// orders that never had a builder set) performs no CPI and does not
+    /// need it, so it may be omitted.
+    pub builder_user: Option<AccountLoader<'info, UserHeader>>,
```
crypto1gmt, line 44: nit: `builder_user` had `has_one = store` and `constraint = builder_user.load()?.is_initialized()` on the account itself in #389. Here `is_initialized()` moved into the handler and the store binding is gone, though `UserHeader` still has the `store` field.

Safe today, since the key has to match `order.builder()`. Happy to leave it for a follow-up rather than grow this PR.

crates/programs/idls/gmsol_store.json
```diff
@@ -28627,13 +28929,11 @@
...
           },
           {
-            "name": "padding_1",
-            "type": {
-              "array": [
-                "u8",
-                8
-              ]
-            }
+            "name": "builder_fee_amount",
```
crypto1gmt, line 29227: suggestion: this looks like it is fixing something rather than just regenerating, and that is worth a line in the description.

#407 replaced `padding_1: [u8; 8]` with `builder_fee_amount: u64` in `states/order.rs`, but the IDL kept `padding_1` and put `builder_fee_amount` after `builder_fee_factor`, taking the 8 bytes out of `reserved`:

| | @2328 | @2384 |
|---|---|---|
| `states/order.rs` (unchanged since #407) | `builder_fee_amount` u64 | `reserved[80]` |
| IDL on `main` | `padding_1[8]` | `builder_fee_amount` u64 |
| IDL here | `builder_fee_amount` u64 | `reserved[80]` |

The tail is 136 bytes either way, so the size still matches and `builder` / `builder_fee_factor` happen to land on the same offsets. `builder_fee_amount` is the one that differs. `Client::order()` goes through `ZeroCopy<store_accounts::Order>`, which is a bytemuck cast, so against `main`'s IDL it succeeds silently and the field reads out of `reserved`, i.e. always 0. `order_account_layout` only asserts Rust offsets, and nothing compares the IDL to the struct.

Harmless while accrual is inert, and this PR's SDK needs the fix anyway since `settle_builder_fee` reads that field to pick between settling and the no-op. Mostly flagging it because `main` carries the broken version today, so it might be worth its own `fix:` PR rather than landing behind this review.

programs/store/src/instructions/builder_fee.rs
gmitsj: Good catch, fixed. claim_vault now carries associated_token::mint = final_output_token / associated_token::authority = builder_user, mirroring the escrow account above it. Anchor's optional-account handling unwraps builder_user for the check and errors if it's None while claim_vault is Some. Dropped the two require_keys_eq! in the handler body since the constraint now covers both.

programs/store/src/instructions/builder_fee.rs
gmitsj: You're right, that's not true yet. I reworded the comment to describe CON-40 as the thing expected to guarantee this, rather than asserting it already happens. Didn't add init_if_needed here since that would duplicate work CON-40 is meant to own, happy to revisit if it ends up not carrying it.

programs/store/src/instructions/builder_fee.rs
gmitsj: Added both back on the account constraint (`has_one = store, constraint = builder_user.load()?.is_initialized()`), matching #389, and removed the now-redundant manual is_initialized() check from the handler body.

crates/programs/idls/gmsol_store.json
gmitsj: This branch's IDL is already regenerated from the current Order struct, which has had builder_fee_amount in the right place since #407, so it's not carrying the bug forward. Confirmed by inspecting the regenerated JSON directly. Agreed it's worth flagging as an explicit fix rather than an incidental regen, I'll add a line to the PR description calling it out. Also agree main should probably get a standalone fix rather than only picking it up whenever the next branch happens to regenerate.

programs/store/src/instructions/builder_fee.rs
angelo-75: We usually do not add `init_if_needed` here, because with the anchor version we are using it tends to consume too much stack space. The usual practice here is to validate its existence (since if downstream really wants to guarantee its existence, it can simply insert an idempotent create-ATA instruction). As for the comment, my suggestion is to describe the actual assumptions of this instruction, not the assumptions of other instructions; that is, state that if provided, this instruction requires the ATA to already exist. Whether to add extra cross-references should be deferred to work like [CON-40](https://linear.app/gmtradepro/issue/CON-40/add-the-owner-signed-set-builder-fee-instruction) (or such clarification can simply be skipped at this stage).

Also, the revised comment is not acceptable, because it references a local identifier like [CON-40](https://linear.app/gmtradepro/issue/CON-40/add-the-owner-signed-set-builder-fee-instruction), and it still describes future work (once that future work is done, the comment becomes false again). I suggest revising it directly along the lines above: describe only the actual assumptions, without referring to other instructions or follow-up work.

programs/store/src/instructions/builder_fee.rs
```diff
@@ -0,0 +1,137 @@
...
+    pub token_program: Program<'info, Token>,
+}
+
+/// Settle the builder fee of the given order.
+///
+/// Permissionless and idempotent: a recorded amount of zero, including
+/// orders that never had a builder set, is an explicit no-op that
+/// performs no token transfer, no state update, and no CPI. This makes
+/// it safe to call in any order state, before or after terminal.
+pub(crate) fn settle_builder_fee(ctx: Context<SettleBuilderFee>) -> Result<()> {
```
angelo-75: suggestion: We should follow this repository's new instruction convention here, i.e., implement `SettleBuilderFee::invoke` instead of a free function. Please refer to: [https://github.com/gmsol-labs/gmx-solana/blob/e1380188642cfbf7da1a6443a56b2326e1f11414/programs/store/src/instructions/token_config.rs#L330](<https://github.com/gmsol-labs/gmx-solana/blob/e1380188642cfbf7da1a6443a56b2326e1f11414/programs/store/src/instructions/token_config.rs#L330>)

angelo-75 (dismissed):

programs/store/src/instructions/builder_fee.rs
gmitsj: I used `invoke` rather than `invoke_unchecked` since neither instruction carries an `#[access_control]` gate, following `SetBuilderFeeFactor::invoke` from #409, which is the closest precedent and already on main.

programs/store/src/instructions/builder_fee.rs
gmitsj: Rewritten to describe only this instruction's assumptions:

```
/// Must already exist when provided; this instruction never creates
/// it, so a missing vault makes settlement fail. Only required
/// alongside `builder_user`, under the same condition.
```

No ticket identifier, no other instruction, nothing that goes stale when follow-up work lands. And no `init_if_needed`. thanks for the note on stack consumption with this Anchor version, validating existence and leaving an idempotent create-ATA to the caller is the better split.

angelo-75 (approved):

PR #409, opened 2026-08-14
feat(store): add permissionless set_builder_fee_factor on the user account

crypto1gmt:
## Summary

Adds `set_builder_fee_factor`, a permissionless instruction with which a User Account owner records the
builder fee factor advertised on their **own** account, bounded by the store-level `MaxBuilderFeeFactor`.

The factor is only an advertisement, and nothing is charged here. An order will pay it only once its owner
checkpoints the factor onto that order, where it is re-validated (ADR-0001); that checkpointing instruction
is a separate issue and is not part of this PR. The cap is therefore enforced in both places on purpose:
here for immediate feedback to the builder, and again at checkpoint time because the cap is
runtime-adjustable and may be lowered after a factor was legally set.

Builds on the state carve-outs merged in #407: `UserHeader.builder_fee_factor` and
`FactorKey::MaxBuilderFeeFactor`.

## What's here

- `SetBuilderFeeFactor` follows `set_referrer`'s account shape: `owner: Signer`, `store`, and the `user`
  account constrained by `has_one = owner`, `has_one = store`, `is_initialized()` and the PDA seeds
  `[UserHeader::SEED, store, owner]`. The seeds are what make this safe to expose permissionlessly: the
  account address is derived from the signer, so setting a factor on someone else's account is
  unconstructible rather than merely rejected.
- The handler reads the cap, rejects anything above it with a new `CoreError::BuilderFeeFactorExceedsMaxFactor`,
  writes the factor and emits `BuilderFeeFactorSet` with the user, its owner, and the previous and new
  factors. The previous value comes from the setter itself, so no second account read is needed.
- `UserHeader` gains a `pub(crate)` setter; #407 shipped the getter only.
- New `events/user.rs` for the event. The order-side builder-fee events live in `events/order.rs`, but this
  one describes the user account, so it gets its own module rather than being filed under orders.
- SDK: `UserOps::set_builder_fee_factor`, mirroring `set_referrer`. Synchronous, since there is no hint to
  fetch.
- A `CHANGELOG.md` entry for the instruction and the SDK method.

## On the feature flag

The instruction is deliberately **not** gated by the `BuilderFee` domain flag, and the handler carries a
comment saying so, because the omission is easy to read as an oversight. It records an advertised rate and
nothing charges a fee off it today; a fee can only be charged once the factor is checkpointed onto an order,
so the gate belongs on that instruction rather than this one.

## On the unset cap

The cap is read with `get_factor_by_key(FactorKey::MaxBuilderFeeFactor)`, and a missing key is treated as
`0` rather than raised as `Unimplemented`, so the failure mode is deny-by-default: on a store where the key
was never wired, only opting out (`0`) remains possible and no nonzero factor is configurable. Since #407
always wires the key, that branch is unreachable today rather than merely unlikely, but the choice matters
because the wrong default here is "unbounded" rather than "denied".

The practical consequence is worth stating explicitly: on every existing store the cap reads `0` until a
config keeper raises it, so this instruction is inert until the protocol deliberately opens the mechanism.

## On the checkpoint-immutability criterion

One acceptance criterion states that changing the factor never affects factors already checkpointed on
orders. Nothing in this PR implements it, and that is intentional, as agreed before implementation started.

The advertisement (`UserHeader.builder_fee_factor`) and the checkpoint (`Order.builder_fee_factor`) are
separate fields on separate accounts, and nothing here writes the checkpoint. So this instruction cannot
affect a checkpoint either way, and the invariant cannot be exercised until the checkpointing instruction
exists. It is stated and owned by that issue, where re-checkpointing semantics are defined in full.

## On the IDL

This PR updates `crates/programs/idls/gmsol_store.json`, which feature PRs normally leave to the separate
`chore: update idls` runs. It is unavoidable here: `declare_program!` generates the SDK's `accounts` and
`args` types from that file, so the SDK support this issue asks for does not compile without it, and
`cargo check --workspace --all-features` would fail on CI.

The diff is additive only: 197 insertions, 0 deletions, covering exactly the instruction, the event, its
type and the new error code. No existing entry is modified. Happy to split it out if you would rather the
IDL moved on its own schedule.

## Tests

One case in `tests/anchor_test/user.rs`, walking the whole lifecycle in the order a builder would hit it:

1. A nonzero factor is rejected while the cap is still `0`, which is the state of every existing store.
2. A config keeper raises the cap, opening the mechanism.
3. Setting the factor exactly at the cap succeeds, and the stored value is read back from the account rather
   than inferred from the transaction succeeding.
4. One unit above the cap is rejected with the new error, and the stored factor is asserted unchanged.
5. `0` succeeds, since that is how a builder opts out.
6. A hand-rolled instruction pointing at another owner's User Account fails on the seeds constraint.

The test restores the cap to `0` at the end, since the fixture is shared with the rest of the suite.

## Out of scope

- `set_builder_fee` and anything about checkpointing a factor onto an order.
- Fee computation, accrual and payout.
- Raising the cap on any live store: that stays a config-keeper decision, and until it happens this
  instruction accepts only `0`.

### DIFF
--- a/programs/store/src/instructions/user.rs
+++ b/programs/store/src/instructions/user.rs
@@ -2,9 +2,10 @@ use anchor_lang::prelude::*;
 use gmsol_utils::InitSpace;
 
 use crate::{
+    events::{BuilderFeeFactorSet, EventEmitter},
     states::{
         user::{ReferralCodeBytes, ReferralCodeV2, UserHeader},
-        Seed, Store,
+        FactorKey, Seed, Store,
     },
     CoreError,
 };
@@ -305,6 +306,72 @@ pub(crate) fn cancel_referral_code_transfer(
         code.code,
         code.next_owner(),
     );
+
+    Ok(())
+}
+
+/// The accounts definitions for
+/// [`set_builder_fee_factor`](crate::gmsol_store::set_builder_fee_factor) instruction.
+#[event_cpi]
+#[derive(Accounts)]
+pub struct SetBuilderFeeFactor<'info> {
+    /// Owner.
+    pub owner: Signer<'info>,
+    /// Store.
+    pub store: AccountLoader<'info, Store>,
+    /// User Account.
+    ///
+    /// The seeds bind this account to the signing `owner`, so setting the
+    /// factor on someone else's User Account is unconstructible rather than
+    /// merely rejected. This is what makes the instruction permissionless
+    /// without granting anyone reach beyond their own account.
+    #[account(
+        mut,
+        has_one = owner,
+        has_one = store,
+        constraint = user.load()?.is_initialized() @ CoreError::InvalidUserAccount,
+        seeds = [UserHeader::SEED, store.key().as_ref(), owner.key().as_ref()],
+        bump = user.load()?.bump,
+    )]
+    pub user: AccountLoader<'info, UserHeader>,
+}
+
+pub(crate) fn set_builder_fee_factor(
+    ctx: Context<SetBuilderFeeFactor>,
+    factor: u128,
+) -> Result<()> {
+    // Note that this instruction is deliberately not gated by the
+    // `BuilderFee` domain flag: it only records an advertised rate, and no
+    // instruction charges a fee off it today. A fee can only be charged after
+    // a future `set_builder_fee` checkpoints the factor onto an order, so
+    // gating belongs there rather than here.
+    //
+    // A missing cap is treated as `0` (deny by default), so on a store that
+    // predates the factor key only opting out remains possible. The key is
+    // always wired today, which makes this branch unreachable rather than
+    // merely unlikely.
+    let max_factor = ctx
+        .accounts
+        .store
+        .load()?
+        .get_factor_by_key(FactorKey::MaxBuilderFeeFactor)
+        .copied()
+        .unwrap_or(0);
+    require_gte!(
+        max_factor,
+        factor,
+        CoreError::BuilderFeeFactorExceedsMaxFactor
+    );
+
+    let previous_factor = ctx.accounts.user.load_mut()?.set_builder_fee_factor(factor);
+
+    let event_emitter = EventEmitter::new(&ctx.accounts.event_authority, ctx.bumps.event_authority);
+    event_emitter.emit_cpi(&BuilderFeeFactorSet::new(
+        ctx.accounts.user.key(),
+        ctx.accounts.owner.key(),
+        previous_factor,
+        factor,
+    ))?;
 
     Ok(())
 }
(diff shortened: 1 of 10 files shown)


programs/store/src/instructions/user.rs
```diff
@@ -308,3 +309,69 @@ pub(crate) fn cancel_referral_code_transfer(
...
+    // predates the factor key only opting out remains possible. The key is
+    // always wired today, which makes this branch unreachable rather than
+    // merely unlikely.
+    let max_factor = ctx
+        .accounts
+        .store
+        .load()?
+        .get_factor_by_key(FactorKey::MaxBuilderFeeFactor)
+        .copied()
+        .unwrap_or(0);
```
angelo-75: nit: Although unwrapping to 0 is reasonable, ~~perhaps erroring out directly would leave fewer questions for the reader.~~

programs/store/src/instructions/user.rs
```diff
@@ -308,3 +309,69 @@ pub(crate) fn cancel_referral_code_transfer(
...
+        has_one = owner,
+        has_one = store,
+        constraint = user.load()?.is_initialized() @ CoreError::InvalidUserAccount,
+        seeds = [UserHeader::SEED, store.key().as_ref(), owner.key().as_ref()],
+        bump = user.load()?.bump,
+    )]
+    pub user: AccountLoader<'info, UserHeader>,
+}
+
+pub(crate) fn set_builder_fee_factor(
```
angelo-75: suggestion: The new convention for instructions is to implement the instruction as an associated function of the Accounts struct. Please refer to:<br>[https://github.com/gmsol-labs/gmx-solana/blob/79a77511fe011f1caf8d1b3b993057ba345dea6a/programs/store/src/instructions/token_config.rs#L231](<https://github.com/gmsol-labs/gmx-solana/blob/79a77511fe011f1caf8d1b3b993057ba345dea6a/programs/store/src/instructions/token_config.rs#L231>)

angelo-75 (dismissed):

programs/store/src/instructions/user.rs
angelo-75: Thought about it again. If some future upgrade does end up removing that key (unlikely), unwrapping to 0 might make this instruction a safe migration-off instruction.

crypto1gmt: Agreed, that's the intent. The comment above it already says it: a missing cap means `0`, so on a store that predates the factor key only opting out stays possible. Leaving it as is.

programs/store/src/instructions/user.rs
crypto1gmt: Done in `1626ab9f`. Went with `invoke` rather than `invoke_unchecked` since this one isn't role-gated, matching `callback.rs` and `UpdateOrderV2`. Tell me if you'd rather keep the suffix uniform.

Left the other handlers in `user.rs` as free functions so the diff stays in scope.

angelo-75 (approved):

PR #413, opened 2026-08-17
feat: add the owner-signed claim_builder_fees instruction

gmitsj:
## Summary

Adds the owner-signed `claim_builder_fees` instruction from ADR-0001. Settled builder fees accumulate in a builder's claim ATA (a token account whose authority is their User Account PDA), one per token mint, and since only the program can sign for that PDA, moving those tokens anywhere requires a dedicated instruction. This is that instruction.

Independent of fee activation and settlement. Claim ATAs can already receive tokens the moment they exist (any transfer to that ATA works today, with or without `settle_builder_fee`), and this is the only way to move tokens back out of one, so it's useful and safely mergeable on its own.

## What's implemented

- **`BuilderFeeClaimed` event** (`programs/store/src/events/builder_fee.rs`), recording the builder (User Account), the token mint, the amount, and the destination.
- **`claim_builder_fees` instruction** (`programs/store/src/instructions/builder_fee.rs`), transferring the full balance of a builder's claim vault to an owner-chosen destination token account, then done. No partial claims: the technical decision in the issue is that the claim vault is a pure collection account, so partial claims have no use case.
  - Restricted to the User Account's owner via `has_one = owner` on `user_account`, the same pattern `InitializeReferralCode` already uses elsewhere in this file's neighborhood. No other path can move tokens out of the claim vault, since its authority is the program-controlled User Account PDA, not a role or a keeper.
  - `user_token_controller` is checked only for PDA derivation (`seeds = [SEED, user_account, token_mint], bump`), nothing else, exactly the placeholder the issue asks for: reserving the account slot so real access control can be added behind it later without a breaking change.
  - Not gated by the `BuilderFee` feature flag. Disabling the feature must never freeze funds that already settled.
  - Idempotent: an empty claim vault returns immediately, no CPI, no state update, no event.
- **SDK support** (`crates/sdk/src/client/ops/builder_fee.rs`, `BuilderFeeOps::claim_builder_fees`). No hint/fetch step needed here, unlike `settle_builder_fee`: every account is a pure function of `store`, `token_mint`, `destination`, and the caller's own key, so it's synchronous.
- A `Client::find_user_token_controller_address` helper (`crates/sdk/src/client/mod.rs`), matching the existing `find_user_address` / `find_referral_code_address` wrappers around the raw `pda::` functions.

## Testing

- `cargo build --workspace` / `cargo test -p gmsol-store -p gmsol-sdk --lib --features client` / `cargo clippy -p gmsol-store -p gmsol-sdk --lib --features client`. All clean, no new warnings.
- No colocated Rust unit tests: there's no pure logic here worth extracting, consistent with how the rest of this instruction layer is tested only through `anchor_test`.
- No `anchor_test` integration coverage yet for the owner-only gate or the no-op path. Unlike `settle_builder_fee`'s PR, this one doesn't have accrual-is-inert as an excuse: claim vaults are exercisable today, so "only the owner can claim" isn't provable by inspection alone. Flagging as a gap to close before merge rather than deferring it.

### DIFF
--- a/programs/store/src/instructions/builder_fee.rs
+++ b/programs/store/src/instructions/builder_fee.rs
@@ -3,11 +3,11 @@ use anchor_spl::token::{transfer_checked, Mint, Token, TokenAccount, TransferChe
 use gmsol_model::action::decrease_position::DecreasePositionSwapType;
 
 use crate::{
-    events::{BuilderFeeSet, BuilderFeeSettled, EventEmitter},
+    events::{BuilderFeeClaimed, BuilderFeeSet, BuilderFeeSettled, EventEmitter},
     states::{
         feature::{ActionDisabledFlag, DomainDisabledFlag},
         user::{UserHeader, USER_TOKEN_CONTROLLER_SEED},
-        FactorKey, Order, Store,
+        FactorKey, Order, Seed, Store,
     },
     CoreError,
 };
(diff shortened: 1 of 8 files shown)


programs/store/src/instructions/builder_fee.rs
```diff
@@ -0,0 +1,126 @@
...
+    /// breaking change to this instruction's accounts.
+    #[account(
+        seeds = [
+            USER_TOKEN_CONTROLLER_SEED,
+            user_account.key().as_ref(),
+            token_mint.key().as_ref(),
+        ],
+        bump,
+    )]
+    pub user_token_controller: UncheckedAccount<'info>,
```
crypto1gmt, line 432: issue (blocking): `test programs` is failing on this head, and this field is the cause.

`anchor idl build` rejects it with `Struct field "user_token_controller" is unsafe, but is not documented`. The doc block explains the account well, but the safety check looks specifically for a line starting with `CHECK:`, and there is none here. `cargo build` and `clippy` do not run that check, which is why the commands listed in the description came back clean. Base `02222f4d` is green on all 8 checks, so it is this branch.

Every other unsafe account field in the repo already carries one. I counted 151 `UncheckedAccount` / `AccountInfo` fields across `programs/`, 71 of them in `programs/store`, and this is the only one without a `CHECK:` doc.

Adding the line costs no IDL regeneration: Anchor strips `CHECK:` entries out of the `docs` array, so `gmsol_store.json` stays exactly as it is. `create_deposit.receiver` is the proof, two doc lines in the source and one in the IDL.

programs/store/src/instructions/builder_fee.rs
```diff
@@ -0,0 +1,126 @@
...
+    /// nonexistent vault the same as an empty one. The account can be
+    /// created permissionlessly ahead of time (e.g. via the associated
+    /// token program's idempotent create instruction) to reach the
+    /// genuine no-op path described below.
+    #[account(
+        mut,
+        associated_token::mint = token_mint,
+        associated_token::authority = user_account,
+    )]
+    pub claim_vault: Box<Account<'info, TokenAccount>>,
```
crypto1gmt, line 409: issue (non-blocking): the no-op AC asks this to mirror the settlement semantics, and I do not think it does yet.

`settle_builder_fee` on #408 reaches its no-op by making `builder_user` and `claim_vault` both `Option<...>`, so the zero path works even when the accounts do not exist. Here `claim_vault` is required and carries the ATA constraints, so a builder who has never had a fee settled for this mint gets `AccountNotInitialized` rather than a successful no-op. Anchor's `Account::try_from` returns that for a System-owned account with zero lamports, which is exactly a never-created ATA.

Your doc comment makes the case for it deliberately and the argument stands on its own. What bothers me is the workaround it points at: creating the ATA permissionlessly first means paying rent so that a no-op can succeed.

Making `claim_vault` an `Option` would mirror #408 exactly. Either way this reads as a deviation from the AC as written, so it is probably @snowy_137's call whether the code moves or the AC does.

programs/store/src/instructions/builder_fee.rs
```diff
@@ -0,0 +1,126 @@
...
+/// Restricted to the [`user_account`](ClaimBuilderFees::user_account)'s
+/// owner: no other signer can initiate a claim, and no other instruction
+/// can move tokens out of a claim vault, since its authority is the
+/// program-controlled User Account PDA. Idempotent: a claim vault with a
+/// zero balance is an explicit no-op, performing no token transfer, no
+/// CPI, and emitting no event (the vault must already exist for this
+/// path; see [`claim_vault`](ClaimBuilderFees::claim_vault)). Not gated
+/// by the `BuilderFee` feature flag, since disabling the feature must
+/// never freeze already-settled funds.
+pub(crate) fn claim_builder_fees(ctx: Context<ClaimBuilderFees>) -> Result<()> {
```
crypto1gmt: suggestion: agreed with your own note in the description, and I would keep this inside the PR rather than after it.

The owner-only gate and the no-op are precisely the two criteria that reading the diff cannot demonstrate, and you already ruled out the accrual-is-inert argument that covered #408. One ordering note: the `claim_vault` question above decides the shape of the no-op test, since it changes whether the test passes that account or omits it.

programs/store/src/instructions/builder_fee.rs
```diff
@@ -0,0 +1,126 @@
...
+    #[account(
+        mut,
+        associated_token::mint = token_mint,
+        associated_token::authority = user_account,
+    )]
+    pub claim_vault: Box<Account<'info, TokenAccount>>,
+    /// The destination token account for `token_mint`, chosen by the
+    /// owner. Not required to be an associated token account.
+    #[account(mut, token::mint = token_mint)]
+    pub destination: Box<Account<'info, TokenAccount>>,
```
crypto1gmt, line 415: nit (non-blocking): nothing stops `destination` from being the claim vault itself.

spl-token short-circuits a self-transfer to `Ok(())` after validation, without touching balances, so the call would emit `BuilderFeeClaimed` with the full vault balance as `amount` while the vault still holds it. No funds move and nothing breaks on chain, but anything indexing the event stream would count a claim that never happened.

A `constraint = destination.key() != claim_vault.key()` rules it out.

crypto1gmt (changes requested):
`test programs` is red on this head, and the cause is inline below. Base `02222f4d` is green on all 8 checks, so it comes from this branch rather than from main.

On the acceptance criteria: owner-only claiming, the full-balance transfer, the UserTokenController derivation check, the feature-flag exemption, the event and the SDK op all look covered to me. The no-op one is the only place where I think the code and the AC text disagree, comment inline on `claim_vault`.

One thing worth a line in the description: the IDL delta here also fixes the `padding_1` / `builder_fee_amount` drift I flagged on #408, so a reader can tell it apart from unrelated regeneration churn.

programs/store/src/instructions/builder_fee.rs
gmitsj: Fixed, and you were right about the IDL too.

My initial fix wrote the `CHECK:` rationale as a three-line paragraph. Anchor strips only the single line that starts with CHECK:, so the two continuation lines survived into `gmsol_store.json` as an orphaned fragment, "\bump` constraints below. The account is never read, written, or", with the sentence it belonged to gone.

programs/store/src/instructions/builder_fee.rs
gmitsj: `claim_vault` is now `Option<Box<Account<'info, TokenAccount>>>`, and the handler returns `Ok(())` when it's absent. A builder who has never had a fee settled for a mint can omit the account and get a genuine no-op, no rent paid to reach it

programs/store/src/instructions/builder_fee.rs
gmitsj: Agreed, and done — `test:` cover owner-only and no-op claim_builder_fees paths.

programs/store/src/instructions/builder_fee.rs
gmitsj: added `constraint = destination.key() != claim_vault.key() @ CoreError::InvalidArgument`, following the same shape as exchange/order.rs:572, with the reasoning in the field's doc comment and a case in the new test.

crypto1gmt (approved):

programs/store/src/instructions/builder_fee.rs
angelo-75: I think no-op here should be understood in the narrow sense: it is enough that no transfer happens when the balance is 0. If the claim vault does not even exist, then rejecting outright here is actually more friendly. Changing it to an Option would not really add much to the value of the no-op, because for the SDK implementation it still requires an extra on-chain query to determine whether the claim vault exists (whereas for the non-Option case, if it is None you can simply not include the instruction in the tx). The truly valuable no-op would be deferring the claim vault's type check.

So my thinking is to drop the Option and revert to the original version, and interpret the no-op in the AC as blocking an unnecessary CPI when the claim vault exists. That said, this discussion is still very valuable, thanks!

angelo-75 (dismissed):

programs/store/src/instructions/builder_fee.rs
gmitsj: Agreed, reverted to the required `claim_vault`

angelo-75 (approved):

PR #415, opened 2026-08-18
feat(gmx-solana): Update pyth-solana-receiver-sdk 1.2.0 (pro-compatible)

RoaringKitty-lab:
### Summary
- Bump to pyth-solana-receiver-sdk 1.2.0 with pro-compatible, Anchor 0.32.1, and Solana ~2.2. Use the new Pyth Core program IDs.
- Probe guardian set 0 in xtask and clone that set (plus the matching config) for local tests. The upgraded Wormhole set lives at index 0.
- Write encoded VAAs in chunks instead of a fixed 755-byte split, so shorter upgraded Hermes payloads can post. just pyth-post is a live check against the new programs.


mv-reyes:
Heads up, the test programs job on this branch is failing and it looks like a toolchain sync issue rather than a code problem. Anchor.toml now pins anchor 0.32.1 / solana 2.2.12, but ci-programs.yml still installs 0.31.1 / 2.1.21 (the comment at the top of the workflow says to keep those in sync with Anchor.toml). In the check run on 3efcaff the installed anchor 0.31.1 reads the new Anchor.toml, tries to hand off to an avm-installed 0.32.1, avm is not on the runner, and the build dies with "os error 2" before anything compiles. So this branch currently gets zero program-build signal from CI. Bumping the two env vars in the workflow should fix it. The README prerequisites and the anchor-lang example at README.md:84 still say 0.31.1 too.

Separately, a small migration-completeness note. The vendored external-programs .so files are unchanged but are now loaded at the new program IDs (Anchor.toml test.genesis and start_localnet.sh). wormhole.so is byte for byte the old mainnet deployment (1,355,328 bytes, matches the old programdata account exactly), while the new deployment is 655,960 bytes. I could not find any behavioral difference that matters, guardian expiry and VAA layout are the same, so tests are not lying today, but the local validator is executing old bytecode at the new addresses, which feels like the kind of thing that bites six months from now. Might be worth re-vendoring from the new deployments while this PR is already touching everything.

PR #416, opened 2026-08-19
feat(store): add the owner-signed set_builder_fee instruction

crypto1gmt:
## Summary

Adds `set_builder_fee`, the owner-signed instruction that checkpoints a builder and its advertised fee
factor onto one of the owner's own pending position orders. This is the counterpart to #409: that one
lets a builder **advertise** a rate on its User Account, this one is how a rate actually becomes owed
on a specific order.

Nothing is charged here either. The checkpoint is what the charging pipeline will later read; execution
and settlement are separate issues.

## What's here

- `SetBuilderFee` copies `UpdateOrderV2`'s account shape for the two authority checks: `owner: Signer`
  plus the order constrained by `header.owner == owner` and `header.action_state()?.is_pending()`. Only
  the owner may attach a builder, and only while the order can still be affected by it.
- The builder is passed as its User Account, constrained by `has_one = store` and `is_initialized()`.
  It is deliberately **not** seed-derived: the builder's wallet is not a party to this transaction, so
  there is no key to derive from, and belonging to this store is the whole of what can be checked.
- The handler requires the passed `expected_factor` to equal what the builder currently advertises,
  exactly. Anything looser lets a builder raise its rate between the owner deciding and the transaction
  landing.
- The store cap is re-checked here, not only at advertise time.
- The claim vault (the builder's ATA for the fee token) must already exist, and the per-user-per-token
  controller PDA is validated by its seeds.
- `Order::set_builder_fee` is the first and only writer of `Order.builder` / `Order.builder_fee_factor`;
  #407 added the fields and getters only.
- `BuilderFeeSet` carries the replaced builder and factor alongside the new ones, so an indexer can
  reconstruct the sequence of checkpoints from the event stream alone.
- Three new `CoreError` variants, appended so existing codes stay stable (6134-6136). The fourth is gone: #406 landed `BuilderFeeSwapTypeNotAllowed` under the same name for the same rule, so this branch uses #406's variant at its own check site instead of adding a duplicate. The overlap flagged in [the comment below](https://github.com/gmsol-labs/gmx-solana/pull/416#issuecomment-5352058505) is resolved by the rebase rather than still pending.

## On the SDK shape, which differs from #408 and #413

Those two put their builder-fee SDK code only in `client/ops/builder_fee.rs`. This one goes in through
the atomic-group builders layer (`builders/order/set_builder_fee.rs`), with the client op reusing it
via `store_program_for_builders`. That is the shape asked for when the issue was scoped, so the
difference is deliberate rather than an oversight; this is just the first of the three to use it.

There is no npm binding, also by decision. The Rust SDK is what the end-to-end tests and the app
project use.

## On the order kind predicate

AC4 says only user-initiated position orders may carry a builder fee, which excludes swaps and the
keeper-initiated kinds. The natural spelling of that, `is_increase_position() || is_decrease_position()`,
is **wrong**: `is_decrease_position()` also matches `Liquidation` and `AutoDeleveraging`, the two kinds
ADR-0001 excludes because the builder there would be the Keeper.

So this adds `OrderKind::is_user_initiated_position` in `gmsol-utils`, spelled as an explicit list of
five kinds rather than composed from the existing two. One of its unit tests asserts that the composed
form **does** accept Liquidation and ADL, so the trap cannot quietly come back.

## On self-checkpointing being allowed

There is no `require!(builder != owner's own User Account)`, and the account carries a comment saying
why. A fresh User Account advertises `0`, and checkpointing one is the only way to clear a checkpoint,
since strict equality means clearing otherwise needs a builder that advertises zero. Forbidding
self-checkpointing reads obviously correct and would delete the only revocation path, leaving an owner
bound to a builder for the life of the order. Flagging it explicitly because it is the kind of thing an
auditor may still want to argue about.

## On the feature flag

Gated on `DomainDisabledFlag::BuilderFee` paired with `ActionDisabledFlag::Default`, and this is the
only builder-fee instruction behind the flag. Settlement, claiming and execution stay ungated on
purpose: disabling the feature should stop new orders from taking on a fee, not freeze a fee already
owed. `Default` is the action because the flag guards a mechanism rather than a step in an order's
lifecycle, and `display_feature` renders it as a bare domain name with no action suffix.

## On the cap being enforced twice

#409 already rejects an advertised factor above `MaxBuilderFeeFactor`, so re-checking here looks
redundant. It is not: the cap is runtime-adjustable, and this is the moment the rate becomes payable
rather than merely advertised. A rate advertised legally at 1% must stop being attachable once the cap
drops to 0.5%. A missing cap key reads `0`, i.e. deny by default, matching the advertising side.

## On the IDL

This PR updates `crates/programs/idls/gmsol_store.json`, which feature PRs normally leave to the
separate `chore: update idls` runs. Unavoidable here for the same reason as #409: `declare_program!`
generates the SDK's `accounts` and `args` types from that file, so the SDK support this issue asks for
does not compile without it.

The diff is purely additive: one instruction (`set_builder_fee`), one event type (`BuilderFeeSet`), and
three error variants at codes 6134-6136. No existing instruction, account, type or error code changes.

An earlier revision of this PR also carried two corrections to pre-existing divergences, `Order`'s field
order and a stale doc string on `final_output_token_escrow`. Both have since landed on main through
#408 and #406, so the rebase drops them: regenerating against `041c72a0` reproduces main's existing
entries unchanged.

## One doc fix outside the feature

`OrderTokenAccounts::final_output_token()` still said "Only available for decrease and swap orders".
That stopped being true with #405, and it is wrong in the direction this instruction depends on, since
AC7 reads the field for increase orders. #405's init is opt-in there rather than unconditional, so the
new wording is that the field is always set for swap and decrease orders, and set for increase orders
only when the creation request supplied the escrow. It is a method doc, so unlike the stale string
above it does not reach the IDL and nothing was regenerated for it.

## Tests

Three cases in `tests/anchor_test/order.rs`:

- **`set_builder_fee`** walks the lifecycle: the builder advertises, a factor one unit off in either
  direction is rejected, a checkpoint signed by anyone but the owner is rejected, the checkpoint lands
  and is read back from the account, an `update_order_v2` in between leaves it untouched, checkpointing
  the owner's own zero-advertising account clears it, and lowering the cap under the advertised rate
  makes the same checkpoint stop working.
- **`set_builder_fee_rejects_ineligible_orders`** covers a swap order, an increase order created
  without the final-output escrow (both at the SDK layer, which refuses to build the instruction, and
  at the program layer when a mint is forced through anyway), a builder with no claim vault, and a
  hand-rolled instruction carrying the wrong controller PDA. It also covers the feature flag: with
  `BuilderFee` disabled, a checkpoint that would otherwise be accepted comes back `FeatureDisabled`, and
  the same call is then repeated with the flag restored and asserted to land. Without that second call
  the assertion would prove nothing, since every other case in the test rejects for a reason of its own.
- **`set_builder_fee_rejects_collateral_to_pnl_swap`** checks that a nonzero factor on a
  `CollateralToPnlToken` decrease order is rejected while zero is still accepted on that same order.

**Liquidation and ADL are covered by unit tests rather than anchor tests, deliberately.** Neither kind
can be created (`create_order` matches the allowed kinds and falls through to `OrderKindNotAllowed`),
and a liquidation is built inside the transaction that executes it, so there is no pending order of
either kind anywhere to point the instruction at. The predicate's own unit tests carry that case
instead, including the one asserting the composed form would admit them.

The test fixture grew `Deployment::with_builder_fee_cap`. `MaxBuilderFeeFactor` is a single global on
a store shared by every test in the binary, they run concurrently, and #409's test already moves it, so
the tests that need it raised had to be serialized against that one and against each other. #409's test
now takes the same lock for its whole body, since its first assertion ("a nonzero factor is rejected
while the cap is zero") only holds while nobody else has it raised. The feature flag is a second store
global of the same shape, so the disabled-feature window reuses that same lock rather than adding a
second primitive, and the accessor is named `lock_builder_fee_globals` to say so. That works because
every test that calls `set_builder_fee` inside a raised-cap window already holds it, and the only test
that calls it outside one is the window's own caller.

## Out of scope

- Charging, accrual, settlement and payout of the fee.
- Any npm SDK binding.
- Raising `MaxBuilderFeeFactor` on any live store, which stays a config-keeper decision. Until it
  happens this instruction accepts only a zero factor.


mv-reyes:
was reading the fee flow on this stack and i think the decrease path has a routing problem. in ops/order.rs the fee is netted before transfer_out (~line 2005), so once the settle todos land, settle_builder_fee pays the recorded fee out of the user's post-fee escrow. user gets charged twice and the fee amount sits stranded in the market vault.

the increase path does it right, it routes the fee into the transfer bucket at 1686. decrease just misses the same step. worth fixing before the todos get wired rather than after.

unrelated but while im here: filed #417, your latest npm release shipped as an empty tarball

crypto1gmt:
you're right, and my first read of this was wrong.

i checked your line numbers against this branch and main, where `ops/order.rs` isn't touched at all, so they landed in the wrong functions and i wrote it off. the charging code is in #406, which i hadn't opened. on that branch `execute_increase_position` is 1596-1792 and `execute_decrease_position` is 1793-2236, so 1686 and 2005 are exactly where you said they were.

the asymmetry is real too. increase routes the fee into the bucket at 1682, decrease subtracts it from `output_amount` at 1974 and then routes only the remainder at 2001. both TODOs feed `order.builder_fee_amount`, and #408 pays the builder out of the escrow, so decrease ends up charging twice.

raising it on #406 since that's where the fix goes. good catch.

re #417, that one's ours, looking at it separately.

crypto1gmt:
Flagging how this overlaps the other builder fee PRs, since it will not merge cleanly alongside #406, #408 and #413 and the overlap is structural rather than accidental.

`programs/store/src/instructions/builder_fee.rs`, `programs/store/src/events/builder_fee.rs` and `crates/sdk/src/client/ops/builder_fee.rs` are each **created** by #416, #408 and #413 alike, one instruction per PR. Whichever lands first owns those files and the other two become edits to them.

The sharper one is `CoreError`. #406 and this PR both append `BuilderFeeSwapTypeNotAllowed` at the same point, with different message text:

- here: `"builder fee is not allowed for this decrease position swap type"`
- #406: `"this decrease position swap type is not allowed together with a builder fee"`

Merged naively that is a duplicate variant and will not compile. The error is genuinely wanted on both sides, since this PR rejects the swap type at checkpoint time and #406 rejects it during execution, so the resolution is to keep one declaration and let both call sites share it rather than to rename either.

Because Anchor numbers errors by declaration order, whichever of the two lands second also has to regenerate the checked-in IDL. The codes here assume `origin/main`'s tail, and #406 inserts three variants ahead of them.

None of this is resolvable from one branch, so I have left it alone rather than guessing at an ordering. Which of the two would you like to land first? Happy to rebase onto it.

crates/sdk/src/client/ops/builder_fee.rs
```diff
@@ -102,4 +121,47 @@ impl<C: Deref<Target = impl Signer> + Clone> BuilderFeeOps<C> for crate::Client<
...
+                    .ok_or_else(|| {
+                        crate::Error::custom(
+                            "the order's final output token is uninitialized, so it cannot carry a builder fee",
+                        )
+                    })?;
+                SetBuilderFeeHint::builder()
+                    .final_output_token(final_output_token)
+                    .build()
+            }
+        };
```
angelo-75: suggestion: Actually there is a standard interface here: `FromRpcClientWith`, which is specifically used for constructing the hint.

programs/store/src/instructions/builder_fee.rs
```diff
@@ -137,3 +142,195 @@ impl SettleBuilderFee<'_> {
...
+                    DecreasePositionSwapType::CollateralToPnlToken
+                )
+            {
+                return err!(CoreError::BuilderFeeSwapTypeNotAllowed);
+            }
+
+            // An increase order created without its final output token escrow
+            // has no slot to pay a fee out of. Uninitialized is a real state
+            // rather than a sentinel, so `None` is the whole of the check.
+            let fee_token = order
```
angelo-75, line 317: suggestion: The Order itself cannot guarantee that if the final output token exists then the final output token escrow must exist (this is only a contract at creation time). However, if a buggy implementation causes the final output token to exist while the final output token escrow does not, and we still set a builder fee for that Order, then at `execute_order` the order will revert during the transfer-out stage, making the order unexecutable (but this is not a deadlock, because if execution fails the builder fee amount is 0, which does not prevent cancellation).

Therefore, consider adding a check that the Order does have a final output token escrow configured.

programs/store/src/lib.rs
```diff
@@ -3156,6 +3156,55 @@ pub mod gmsol_store {
...
+    ///   must not have the [`BuilderFee`](crate::states::feature::DomainDisabledFlag::BuilderFee)
+    ///   feature disabled.
+    /// - The [`order`](SetBuilderFee::order) must:
+    ///   - Belong to the `store`
+    ///   - Still be pending execution
+    ///   - Be a user-initiated position order, so swap orders and the
+    ///     keeper-initiated kinds are rejected
+    ///   - Have an initialized final output token, equal to the passed
+    ///     [`final_output_token`](SetBuilderFee::final_output_token)
+    ///   - Not be a decrease order using
```
angelo-75: nit: In fact the code checks all user-initiated position orders, not just decrease orders, which is stricter than what the comment says.

programs/store/src/instructions/builder_fee.rs
```diff
@@ -137,3 +142,195 @@ impl SettleBuilderFee<'_> {
...
+                kind.is_user_initiated_position(),
+                CoreError::BuilderFeeOrderKindNotAllowed
+            );
+
+            // `CollateralToPnlToken` moves the entire collateral-token output
+            // into the pnl token before the receive-token swap runs, so the
+            // final output token bucket a fee is taken from is empty on every
+            // such order. A zero factor is still allowed, since it charges
+            // nothing and is how a checkpoint gets cleared.
+            if factor != 0
```
angelo-75, line 305: nit: It might be worth adding the scenario to further clarify why we don't unconditionally reject `DecreasePositionSwapType::CollateralToPnlToken`, and only reject it when `builder_factor != 0`, even though it is indeed harmless.

tests/anchor_test/order.rs
```diff
@@ -996,3 +1004,598 @@ async fn increase_order_with_mismatched_final_output_token_should_fail() -> eyre
...
+
+    deployment
+        .with_builder_fee_cap(CAP, async {
+            let signature = builder
+                .set_builder_fee_factor(store, CAP)?
+                .send_without_preflight()
+                .await?;
+            tracing::info!(%signature, "the builder advertised its factor");
+
+            // AC2: the checkpoint carries the factor the owner signed for, and
```
angelo-75: nit: Remove internal identifiers like AC2.

tests/anchor_test/order.rs
```diff
@@ -996,3 +1004,598 @@ async fn increase_order_with_mismatched_final_output_token_should_fail() -> eyre
...
+                    .send()
+                    .await
+                    .expect_err("should reject a factor the builder is not advertising");
+                assert_eq!(
+                    gmsol_sdk::Error::from(err).anchor_error_code(),
+                    Some(CoreError::BuilderFeeFactorMismatched.into()),
+                );
+            }
+
+            // AC1: only the order's owner may checkpoint onto it. Signed here by
```
angelo-75: nit: Remove internal identifiers like AC1.

tests/anchor_test/order.rs
```diff
@@ -996,3 +1004,598 @@ async fn increase_order_with_mismatched_final_output_token_should_fail() -> eyre
...
+            assert_eq!(
+                checkpointed.builder, builder_user,
+                "the builder must be recorded on the order"
+            );
+            assert_eq!(
+                checkpointed.builder_fee_factor, CAP,
+                "the advertised factor must be the one recorded"
+            );
+
+            // AC8: updating the order is not a way to change the checkpoint.
```
angelo-75: nit: Remove internal identifiers like AC8.

tests/anchor_test/order.rs
```diff
@@ -996,3 +1004,598 @@ async fn increase_order_with_mismatched_final_output_token_should_fail() -> eyre
...
+            tracing::info!(%signature, "revoked the builder fee");
+
+            let revoked = owner.order(&order).await?;
+            assert_eq!(revoked.builder, owner_user);
+            assert_eq!(
+                revoked.builder_fee_factor, 0,
+                "checkpointing a zero-advertising account must clear the fee"
+            );
+
+            // AC3: the cap is enforced again at checkpoint time, not only when
```
angelo-75: nit: Remove internal identifiers like AC3.

tests/anchor_test/order.rs
```diff
@@ -996,3 +1004,598 @@ async fn increase_order_with_mismatched_final_output_token_should_fail() -> eyre
...
+        .mint_or_transfer_to_user("fBTC", Deployment::DEFAULT_USER, collateral_amount)
+        .await?;
+    deployment
+        .mint_or_transfer_to_user("USDG", Deployment::DEFAULT_USER, swap_amount + 17)
+        .await?;
+
+    let size = 5_000 * MARKET_USD_UNIT;
+    let price = 400_000 * MARKET_USD_UNIT / 10u128.pow(fbtc.config.decimals as u32);
+
+    // AC4: a swap order is not a position order, so it can carry no builder.
```
angelo-75: nit: Remove internal identifiers like AC4.

tests/anchor_test/order.rs
```diff
@@ -996,3 +1004,598 @@ async fn increase_order_with_mismatched_final_output_token_should_fail() -> eyre
...
+
+    let signature = client
+        .close_order(&swap_order)?
+        .build()
+        .await?
+        .send()
+        .await?;
+    tracing::info!(%swap_order, %signature, "cancelled the swap order");
+
+    // AC7: an increase order created without the escrow has no final output
```
angelo-75: nit: Remove internal identifiers like AC7.

tests/anchor_test/order.rs
```diff
@@ -996,3 +1004,598 @@ async fn increase_order_with_mismatched_final_output_token_should_fail() -> eyre
...
+
+    let signature = client
+        .close_order(&bare_order)?
+        .build()
+        .await?
+        .send()
+        .await?;
+    tracing::info!(%bare_order, %signature, "cancelled the order without the escrow");
+
+    // AC6 needs an order that passes everything else.
```
angelo-75: nit: Remove internal identifiers like AC6.

tests/anchor_test/order.rs
```diff
@@ -996,3 +1004,598 @@ async fn increase_order_with_mismatched_final_output_token_should_fail() -> eyre
...
+            true,
+            collateral_amount,
+        )
+        .prepare_final_output_token_escrow(true)
+        .build_with_address()
+        .await?;
+    let signature = rpc.send().await?;
+    tracing::info!(%order, %signature, "created a fee-eligible limit increase order");
+
+    // AC6, missing claim vault. The keeper's User Account has no fBTC ATA, and
```
angelo-75: nit: Remove internal identifiers like AC6.

tests/anchor_test/order.rs
```diff
@@ -996,3 +1004,598 @@ async fn increase_order_with_mismatched_final_output_token_should_fail() -> eyre
...
+        .await?
+        .send()
+        .await
+        .expect_err("should reject a builder whose claim vault does not exist");
+    assert_eq!(
+        gmsol_sdk::Error::from(err).anchor_error_code(),
+        Some(ErrorCode::AccountNotInitialized.into()),
+    );
+
+    // AC6, the controller PDA. The SDK always derives it, so a mismatch has to
```
angelo-75: nit: Remove internal identifiers like AC6.

tests/anchor_test/order.rs
```diff
@@ -996,3 +1004,598 @@ async fn increase_order_with_mismatched_final_output_token_should_fail() -> eyre
...
+        .anchor_args(args::SetBuilderFee { expected_factor: 0 })
+        .send()
+        .await
+        .expect_err("should reject a controller that is not the derived PDA");
+    assert_eq!(
+        gmsol_sdk::Error::from(err).anchor_error_code(),
+        Some(ErrorCode::ConstraintSeeds.into()),
+    );
+
+    // AC9: the mechanism sits behind `DomainDisabledFlag::BuilderFee`, and this
```
angelo-75: nit: Remove internal identifiers like AC9.

tests/anchor_test/order.rs
```diff
@@ -996,3 +1004,598 @@ async fn increase_order_with_mismatched_final_output_token_should_fail() -> eyre
...
+        .await?;
+    tracing::info!(%order, %signature, "the same checkpoint lands once the feature is back on");
+
+    let signature = client.close_order(&order)?.build().await?.send().await?;
+    tracing::info!(%order, %signature, "cancelled the order");
+
+    Ok(())
+}
+
+/// AC5a: a `CollateralToPnlToken` decrease order can carry no nonzero fee.
```
angelo-75: nit: Remove internal identifiers like AC5a.

angelo-75 (dismissed):

crates/sdk/src/client/ops/builder_fee.rs
crypto1gmt: Good call, switched to it. `SetBuilderFeeHint` now implements `FromRpcClientWith<SetBuilderFee>` in `builders/order/set_builder_fee.rs`, and the client op just calls it, same shape as `UpdateClosedStateHint` in `ops/exchange/market_state.rs`. The uninitialized-token message moved down with it, so both SDK entry points get it now instead of only the op.

One deliberate choice: I call `from_rpc_client_with` directly rather than going through `into_atomic_group_with_rpc_client`, since that one is behind `cfg(client_traits)` and every current use of it in the tree is in a test. The op builds the instruction first and borrows it for the hint, so the explicit-hint path is untouched.

programs/store/src/instructions/builder_fee.rs
crypto1gmt: Agreed, added. The handler now requires `final_output_token.account().is_some()`, with a new `BuilderFeeFinalOutputTokenEscrowNotInitialized` appended at 6137. Your reasoning is in the comment above the check, including the part about it not being a deadlock, since that is the bit that explains why it refuses the checkpoint rather than doing anything cleverer.

No test for it, and that is deliberate rather than an omission. `TokenAndAccount::init` writes the mint and the escrow address together from one token account, so there is no creation path that can produce token-without-escrow today. It is exactly the buggy-future-path guard you describe, so a test would have to fabricate the state rather than reach it. I recorded the absence in the test module's doc comment next to the Liquidation/ADL ones, so the gap is explained where someone would look for the missing case.

Worth noting this is a fourth `CoreError` variant, so the checked-in IDL needed regenerating. Details in the summary comment.

programs/store/src/lib.rs
crypto1gmt: You're right, fixed. The doc now says the check applies to every user-initiated position order rather than just decrease ones.

I also added why the extra breadth is harmless, since that was the other half of the surprise: `decrease_position_swap_type` is only ever written by `init_decrease`, so an increase order always carries `NoSwap` and can never trip the guard. It is unreachable rather than restrictive. That is in the code comment too, not only the entrypoint doc.

programs/store/src/instructions/builder_fee.rs
crypto1gmt: Added. The scenario is revocation: a zero-factor checkpoint is how an owner clears a builder, so gating on `factor != 0` keeps that path reachable on every order whatever its swap type.

If the guard were unconditional, and an order's swap type could ever move after a checkpoint had been written, the existing checkpoint would be stranded, because the only instruction that could clear it would itself be rejected. `update_order_v2` cannot move the swap type today, so this is about not building the trap in rather than a live bug, and I said so in the comment rather than implying the hole exists now.

tests/anchor_test/order.rs
crypto1gmt: Done.

tests/anchor_test/order.rs
crypto1gmt: Done.

tests/anchor_test/order.rs
crypto1gmt: Done.

tests/anchor_test/order.rs
crypto1gmt: Done.

tests/anchor_test/order.rs
crypto1gmt: Done.

tests/anchor_test/order.rs
crypto1gmt: Done.

tests/anchor_test/order.rs
crypto1gmt: Done.

tests/anchor_test/order.rs
crypto1gmt: Done.

tests/anchor_test/order.rs
crypto1gmt: Done.

tests/anchor_test/order.rs
crypto1gmt: Done.

tests/anchor_test/order.rs
crypto1gmt: Done.

crypto1gmt:
All 15 addressed, pushed as `f2262cd3`, `9cde7532` and `6430a2c4`. CI is green, 8/8 on the head.

Two things worth pulling out of the threads.

I removed a twelfth `AC4` you didn't flag, in the doc comment on `set_builder_fee_rejects_ineligible_orders`. `grep` for internal identifiers now comes back empty across the tree, not just in the lines you marked.

The escrow check adds a fourth `CoreError` variant, so the checked-in IDL was regenerated. I diffed it structurally against `main` rather than eyeballing it: purely additive, zero shifted error codes, and no existing instruction, account, type or event changed. The delta is `set_builder_fee`, `BuilderFeeSet` and the four errors at 6134-6137. The description above still says three at 6134-6136, so I'll correct that count.

The push dismissed your approval, so this needs another pass when you have a slot.

angelo-75 (approved):

PR #418, opened 2026-08-24
fix(store)!: prevent over-minting GT on insolvent liquidation/ADL closes

gmitsj:
## Problem

During an insolvent Liquidation or ADL close, `paid_order_and_borrowing_fee_value` could stay non zero even though the matching fees were never collected. The Store program mints GT from that value, so those closes created GT with no fee backing.

The collateral pipeline in `decrease_position/mod.rs` runs in this order.

| # | Stage | Insolvency step |
|---|---|---|
| 4 | `pay_for_funding_fees` | `Funding` |
| 5 | `pay_for_pnl_if_negative` | `Pnl` |
| 6 | `pay_for_fees_excluding_funding` | `Fees` |
| 7 | `pay_for_price_impact_if_negative` | `Impact` |
| 8 | `pay_for_price_impact_diff` | `Diff` |

`CollateralProcessor::process` swallows `InsufficientFundsToPayForCosts` when insolvent close is allowed and returns the report early, so every stage after the failing one is skipped. Liquidation and ADL are the two order kinds that allow it.

When insolvency tripped at stage 4 or 5, stage 6 never ran. Stage 6 held the only code that zeroed `paid_order_and_borrowing_fee_value`, on its partial payment branch. The field therefore kept the value computed up front, and that value flows unguarded into `unchecked_process_gt`.

Reproduced before the fix. A full close after a price crash reported `paid_order_and_borrowing_fee_value == 40_000_000` having collected nothing.

## Fix

Clear the uncollected fees once the pipeline returns, when the reported step precedes the fee step.

Only `Funding` and `Pnl` qualify.

* `Fees` already clears what it failed to collect.
* `Impact` and `Diff` run after the fee step, so their fees were genuinely collected. A blanket `insolvent_close_step.is_some()` check would under mint GT for them.

The decision lives in `insolvency_preempted_fee_collection` so it can be tested on its own. It matches exhaustively rather than comparing by ordering, because the declaration order of `InsolventCloseStep` is `Pnl, Fees, Funding, Impact, Diff`, which does not match pipeline order, and the enum derives no `Ord`. Being in the crate that defines the enum, the exhaustive match also forces a decision if a sixth step is ever added.

Fixed in the model rather than at the GT call site, so every consumer of the report sees a consistent value and the ordering decision is not duplicated where two copies could drift.

## Testing

`cargo test -p gmsol-model` passes 36 tests and `cargo test -p gmsol-store --lib` passes 32. `cargo fmt --check` is clean and clippy reports nothing on the touched files. The regression test was confirmed to fail when the fix is reverted.

No IDL regeneration is needed, since this changes values rather than the shape of `DecreasePositionReport` or `PositionDecreased`.

## Coverage gap

There is no end to end anchor test for the fixed path. Driving an on chain position into `Pnl` step insolvency requires the index price to move far enough that PnL exceeds collateral, and `execute_with_pyth` sources prices from live Hermes feeds with no override. That is why the existing `liquidation` test forces liquidatability by raising `MinCollateralFactorForLiquidation` at the live price instead, which is a solvent close that never reaches this path.

Covering it on chain would need new harness capability, such as a price override or a mock feed. The model tests cover the logic, and commit 4 pins the store side link that turns a zero value into zero GT.

`Funding` step insolvency is covered by the decision table but not end to end either. It shares a branch with `Pnl`, and driving a close insolvent on funding fees alone would require advancing funding state until it exceeds collateral.

### DIFF
--- a/crates/model/src/action/decrease_position/collateral_processor.rs
+++ b/crates/model/src/action/decrease_position/collateral_processor.rs
@@ -557,10 +557,10 @@ where
                         // Empty the fees since the amount was entirely paid to the pool instead of for fees.
                         // It is possible for the txn execution to still complete even in this case
                         // as long as the remaining_cost is still zero.
-                        fees.clear_fees_excluding_funding();
-
-                        // Also, clear the `paid_order_and_borrowing_fee_value` for the same reason.
-                        fees.set_paid_order_and_borrowing_fee_value(Zero::zero());
+                        //
+                        // This also clears `paid_order_and_borrowing_fee_value`, which must not
+                        // outlive the fees it accounts for.
+                        fees.clear_uncollected_fees_excluding_funding();
                     }
                     Ok(())
                 },
--- a/crates/model/src/action/decrease_position/mod.rs
+++ b/crates/model/src/action/decrease_position/mod.rs
@@ -6,8 +6,8 @@ use crate::{
     params::fee::PositionFees,
     pool::delta::PriceImpact,
     position::{
-        CollateralDelta, Position, PositionExt, PositionMut, PositionMutExt, PositionStateExt,
-        WillCollateralBeSufficient,
+        CollateralDelta, InsolventCloseStep, Position, PositionExt, PositionMut, PositionMutExt,
+        PositionStateExt, WillCollateralBeSufficient,
     },
     price::{Price, Prices},
     BorrowingFeeMarketExt, PerpMarketMut, PoolExt,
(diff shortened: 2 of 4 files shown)


angelo-75 (approved):

angelo-75:
There is a breaking change here: the fees in the Trade Event and Report were previously still recorded on an insolvent close that occurs during the funding and pnl steps, but are no longer recorded now.

PR #419, opened 2026-08-25
docs(store): document the builder fee invariants (CON-42)

crypto1gmt:
Documents the properties the builder fee mechanism upholds, as doc comments in the repo rather than as a document elsewhere, so the argument sits next to the code it is about and versions with it.

## What is here

A module-level doc block on `states/order.rs` stating seven invariants and, for each, the argument for why the mechanism upholds it: conservation, coverage, authorization, boundedness, delivery, isolation, liveness. Plus a closing section on the two cases that cannot be exercised on chain and why.

Two existing comments already referred to "the charging invariant" and "a broken invariant" with nothing to point at. Both now name the **coverage** invariant and refer to the catalogue, folding "charging" into it so there is one name per property.

`UpdateOrderParams` gains a doc comment saying that adding a field there is a security decision, and a test pins the field set: the destructuring is exhaustive with no rest pattern, so a new field fails the build rather than silently widening what an update can reach. That matters because a fee is charged at the factor checkpointed on the order and taken from a bucket the decrease position swap type can empty, so neither value may become updatable without revisiting what a checkpoint means.

## Placement

Resolves CON-42.

The catalogue is a module doc on `states/order.rs`, and each enforcement site carries one line naming the invariant it upholds. The split was agreed on the issue.

The second commit adds that per-site layer across 14 sites, covering all seven invariants:

| Invariant | Sites |
|---|---|
| conservation | increase-path record + route, decrease-path record |
| coverage | `clamp_builder_fee_amount`, `charge_builder_fee_on_collateral_increment`, decrease-path record, the close guard |
| authorization | the factor-equality check in `set_builder_fee`, `UserHeader::set_builder_fee_factor` |
| boundedness | `compute_builder_fee_amount`, both cap enforcement points, `Factors::max_builder_fee_factor` |
| delivery | the settlement builder-equality check, the close guard |
| isolation | `compute_builder_fee_amount`, `estimate_builder_fee_for_collateral_withdrawal` |
| liveness | the feature-flag gate |

Convention: name the invariant, with the `crate::states::order` pointer on the first annotation in each file rather than repeated at all fourteen.

## The catalogue says charging is not wired yet

When this branch was first written the execution call sites still passed a literal zero factor, so the charging invariants held vacuously and the doc block said so, naming the change that would make them live. That change is #421, which has since landed, so the doc block now states that charging is wired and the invariants are load-bearing. Rebasing onto it is what surfaced the discrepancy.

## Scope notes

- The IDL carries `docs` arrays, so any annotated item that appears in it is updated too. The delta is three entries across the two commits, `UpdateOrderParams`, `BuilderFeeSettled.settled_amount` and `Factors.max_builder_fee_factor`, grafted onto the checked-in file rather than copied wholesale; nothing else in the generated IDL differed.
- Nothing outside doc comments, one test and that IDL delta is touched. No behaviour changes.

### DIFF
--- a/crates/programs/idls/gmsol_store.json
+++ b/crates/programs/idls/gmsol_store.json
@@ -27028,7 +27028,10 @@
             "docs": [
               "Max builder fee factor. Reads as `0` on existing stores until a",
               "CONFIG_KEEPER explicitly raises it via `insert_factor`, so no",
-              "non-zero builder fee factor can be configured by default."
+              "non-zero builder fee factor can be configured by default.",
+              "",
+              "The cap the boundedness invariant is stated against; see the builder",
+              "fee invariants on [`crate::states::order`]."
             ],
             "type": "u128"
           },
--- a/crates/programs/idls/gmsol_store.json
+++ b/crates/programs/idls/gmsol_store.json
@@ -25040,8 +25040,9 @@
             "name": "settled_amount",
             "docs": [
               "The amount actually transferred, clamped to the escrow's balance.",
-              "A value below `recorded_amount` signals a broken invariant, since",
-              "the escrow is expected to always cover the recorded amount."
+              "A value below `recorded_amount` breaks the coverage invariant (see",
+              "the builder fee invariants on [`crate::states::order`]), which has",
+              "the escrow always covering the recorded amount."
             ],
             "type": "u64"
           }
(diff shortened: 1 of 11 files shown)


crypto1gmt:
Superseded by #431, re-opened from a branch in this repository so the CI jobs can reach the repo secrets (a `pull_request` run from a fork gets none, which is why `test programs` never really executed here). Same commit `a1ba58a8`, nothing rewritten. Closing this one to keep the review board clean.

PR #422, opened 2026-08-28
fix(store): reject leaving a disabled virtual inventory from an unassociated market

crypto1gmt:
`leave_disabled_virtual_inventory` decrements a virtual inventory's `ref_count` without checking that the market is associated with it. The `if / else if` clears the market's reference only when the address matches one of its two slots, while the decrement sits outside the branch and always runs.

A market that never joined can therefore spend a reference it does not hold. The counter reaches zero early, `close_virtual_inventory` (which gates only on `ref_count == 0`) succeeds, and markets still referencing the account are left pointing at nothing. `RevertibleMarket::new` resolves that reference against the accounts supplied with the transaction and returns `InvalidArgument` when it is absent, so those markets then fail deposits, withdrawals, orders and swaps.

## Change

Return `PreconditionsAreNotMet` when the address matches neither reference. That is the same error the two healthy leave paths already raise for a missing reference, so this adds no error code and needs no IDL update.

The condition is "associated through either reference" rather than a copy of the swaps wrapper's `require_keys_eq!`. This instruction serves both virtual inventory kinds through the same branch, so validating a single reference would break the positions case.

## Tests

Five unit tests next to the model method (swaps reference cleared, positions reference cleared, unassociated market rejected, market holding no reference rejected, enabled virtual inventory rejected), plus an integration test in `tests/anchor_test/virtual_inventory.rs`. There were no virtual inventory tests of any kind before this.

The integration test creates its own two markets and its own virtual inventory at an unused index. The fixture markets are shared with tests that run concurrently, and a market joined to a virtual inventory must carry that account in every transaction touching it, so borrowing one would make unrelated tests fail intermittently.

Both halves were checked against a build with the fix reverted, so they demonstrably catch the bug rather than merely pass: the two rejection unit tests fail, and in the integration test the unassociated market's `leave_disabled` succeeds and returns a signature. Stating that here because `CI for Programs` is currently red across the repository for an unrelated reason (the recent Pyth Core upgrade withdrew the Wormhole guardian set accounts the validator clones), so the integration half will not run on this PR.

## Notes

The existing recovery procedure for an already-damaged virtual inventory is unaffected: the markets it has leave are genuinely associated, so they pass the new check.

No repair step is included because there is nothing to repair. Enumerating every virtual inventory and every market of the store program on devnet (34 markets, 3 virtual inventories) and mainnet-beta (101 markets, 2 virtual inventories) shows `ref_count` matching the number of markets actually referencing each account in every case, and no market referencing an account that no longer exists.

### DIFF
--- a/programs/store/src/states/market/mod.rs
+++ b/programs/store/src/states/market/mod.rs
@@ -564,6 +564,12 @@ impl Market {
 
     /// Leave a disabled virtual inventory.
     ///
+    /// This market must currently be associated with the given [`VirtualInventory`],
+    /// through either of its two references. Otherwise the `ref_count` would be
+    /// decremented on behalf of a market that never joined, which would let the
+    /// virtual inventory reach zero and be closed while other markets still
+    /// reference it.
+    ///
     /// # CHECK
     /// - The address and the provided [`VirtualInventory`] must match.
     pub fn leave_disabled_virtual_inventory_unchecked(
@@ -580,6 +586,9 @@ impl Market {
             self.virtual_inventory_for_swaps = DEFAULT_PUBKEY;
         } else if self.virtual_inventory_for_positions() == Some(address) {
             self.virtual_inventory_for_positions = DEFAULT_PUBKEY;
+        } else {
+            msg!("the market is not associated with this virtual inventory");
+            return err!(CoreError::PreconditionsAreNotMet);
         }
 
         virtual_inventory.leave_unchecked(Delta::new(None, None))?;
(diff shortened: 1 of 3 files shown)


crypto1gmt:
Superseded by #432, re-opened from a branch in this repository so the CI jobs can reach the repo secrets (a `pull_request` run from a fork gets none, which is why `test programs` never really executed here). Same commit `dfb976b0`, nothing rewritten. Closing this one to keep the review board clean.

PR #424, opened 2026-09-02
fix(store): drop the builder fee recorded by a discarded execution

crypto1gmt:
`Order` is not revertible. `perform_execution` takes it as a plain `&mut Order`, unlike `RevertiblePosition` and the markets, so anything written to it survives a soft failure. The builder fee is recorded part-way through execution, and when a later step fails with `cancel_on_execution_error` set, the `TransferOut` that was supposed to fund that fee is replaced by `Default::default()` while the record stays. The order is then cancelled, and every close path refuses it until `settle_builder_fee` has run against an escrow nothing was routed into.

Introduced by #421, which replaced four hardcoded zero fee factors with the order's checkpointed one; before that `record_builder_fee(0)` made the path unreachable.

### The fix

Snapshot `builder_fee_amount` before `perform_execution` and restore it in the soft-failure arm, next to where `transfer_out` is discarded.

Keyed on the outcome rather than on an ordering. Moving the record after "the last fallible step" is not sufficient: `perform_execution` still calls `write_to_event` and `update_with_transfer_out` after `execute_increase_position` returns, and any such list has to be re-derived every time the function changes.

Restores rather than zeroes, because `record_builder_fee` accumulates via `checked_add` (pinned by `recorded_builder_fee_amounts_accumulate`). Restoring is correct without depending on an order being executable at most once.

A hard failure needs no handling: the runtime reverts the account write.

### Why the snapshot window is safe

If anything could zero the field between snapshot and restore, restoring would resurrect an amount settlement had already paid out. The only callback CPI inside `ExecuteOrderOperation` is in `handle_executed`, which runs after the match arm; the rest of the window is token-program transfers and the store's own `emit_cpi`. The only two writers of the field are `record_builder_fee` and the settlement zeroing.

### Test

`soft_failed_execution_records_no_builder_fee` checkpoints a fee onto a `MarketDecrease` with an unreachable `min_output_amount`, which fails `validate_decrease_output_amounts` after the fee is recorded and with `should_throw_error` left false, so the failure lands in the soft-failure arm. It asserts the order is `Cancelled`, that the recorded amount is back to its pre-execution value, and that the order closes without settling anything.

The decrease path is used because that minimum is a parameter the test controls; the increase path reaches the same arm only through a market condition. It executes with the bundled close disabled, so the assertion lands on the recorded amount itself rather than on the close guard tripping.

### Note for reviewers

The regression test has now been run locally, both ways. The only difference between the two runs is the two-file fix in `programs/store`:

| code | result |
| --- | --- |
| without the fix | `FAILED. 0 passed; 1 failed`, 105.73s |
| with the fix | `ok. 1 passed; 0 failed`, 113.26s |

It fails on its own assertion rather than on infrastructure:

```
assertion `left == right` failed: a discarded execution must leave the recorded builder fee untouched
  left: 13
 right: 0
```

So a discarded execution recorded a builder fee of 13 against an order whose principal had been refunded, and the fix brings it to 0.

**CI status.** [#423](https://github.com/gmsol-labs/gmx-solana/pull/423) and [#425](https://github.com/gmsol-labs/gmx-solana/pull/425) are both merged, and this branch is rebased on top of them, so the validator starts and the suite runs.

One gap is left, and it is not fixable from this PR: the `test programs` job has no `PYTH_API_KEY` secret, so every oracle test gets a 401 from Hermes, **including the regression test added here**. Only an admin can add the secret. The rest of the suite runs normally.

Locally, with the key set, the test was run both ways and the results are in the table above.

One more thing worth flagging: the comment at `programs/store/src/instructions/builder_fee.rs:336` already asserts that "a failed execution charges a zero fee". That was false on `main` and is true again with this change.

### DIFF
--- a/programs/store/src/ops/order.rs
+++ b/programs/store/src/ops/order.rs
@@ -951,6 +951,19 @@ impl ExecuteOrderOperation<'_, '_> {
         let mut should_throw_error = false;
         let prices = self.prices()?;
         let discount = self.validate_and_get_order_fee_discount()?;
+
+        // The builder fee is recorded onto the order part-way through
+        // execution, and the order account is not revertible: unlike the
+        // position and the markets, it is passed down as a plain `&mut Order`,
+        // so a soft failure below keeps whatever was written to it while
+        // discarding the `TransferOut` that was supposed to fund the fee.
+        //
+        // Snapshotting here keys the record to the outcome rather than to an
+        // ordering of the fallible steps. Moving the record after the last one
+        // would fix the paths that exist today and silently rot the moment a
+        // new fallible step is added after it.
+        let builder_fee_amount_before = self.order.load()?.builder_fee_amount();
+
         let res = match self.perform_execution(&mut should_throw_error, prices, discount) {
             Ok((should_remove_position, mut transfer_out, should_send_trade_event)) => {
                 transfer_out.set_executed(true);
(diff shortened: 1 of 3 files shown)


crypto1gmt:
Superseded by #433, re-opened from a branch in this repository so the CI jobs can reach the repo secrets (a `pull_request` run from a fork gets none, which is why `test programs` never really executed here). Same commit `935e1495`, nothing rewritten. Closing this one to keep the review board clean.

PR #425, opened 2026-09-03
fix(sdk): handle short Pyth VAAs, and refresh the vendored Wormhole program

crypto1gmt:
Pyth rotated Wormhole guardian set 0 on 2026-08-26 and now publishes 292-byte VAAs signed by 3 of 5 guardians. Three things here broke as a result: every oracle-driven test fails, and one of the three also hits anyone using the SDK to post Pyth prices.

### 1. `pull_oracle_impl.rs` panics on any VAA shorter than 755 bytes

This one is shipped code, not test-only. `&vaa[0..VAA_SPLIT_INDEX]` sliced unconditionally at 755, so every price delivery through `PythPullOracleWithHermes` now panics:

```
range end index 755 out of range for slice of length 292
```

The split exists because a VAA *can* exceed one transaction, which is not a property of every VAA: the length scales with the guardian signature count. It now splits only when there is something to split, and a short VAA goes out in a single `write_encoded_vaa`.

### 2. `external-programs/wormhole.so` was built in 2024 and answers `NoQuorum` (0x1973)

The tests load Wormhole from this vendored binary via `[[test.genesis]]`, not from the cluster, so the deployed program never gets a say. The committed copy predates the change that made quorum `n / 2 + 1`: it demands 4 signatures from the 5-key set while Pythnet publishes 3.

Refreshed from the deployed program. Independently verifiable:

```
solana program dump HDwcJBJXjL9FpJ7UBsYBtaDjsBUhuLCUYoz3zr8SWWaQ wormhole.so -u d
sha256sum wormhole.so
ca237355a04cc5ceeb788853833ae46c0cb8b0a728625e8f9bb9bc05d1e9c619  # 1356352 bytes, matches this PR
```

### 3. The fixture never read `PYTH_API_KEY`

Hermes requires a key now, and `setup.rs` built the client with `Default::default()`, whose `Default` hardcodes `api_key: None`. It now uses `Hermes::from_default_envs()`, falling back to the previous behaviour when the variable is unset, so contributors without a key are unaffected.

### Evidence

CI cannot show any of this yet. `test programs` is red on `main` itself, and has been since the 2026-08-26 rotation: `Anchor.toml` still clones guardian set 7 (`6GaHgiaQ...`), which that rotation withdrew, so `solana-test-validator` refuses to start with `Account 6GaHgiaQ... not found` and the job fails before a single test runs. [#423](https://github.com/gmsol-labs/gmx-solana/pull/423) fixes exactly that, and this branch is cut from `main`, so it inherits the failure. Measured locally instead, against `anchor_test::order::balanced_market_order`, changing only the binary:

| wormhole.so | result |
| --- | --- |
| vendored (2024) | fails on `verify_encoded_vaa_v1`, `custom program error: 0x1973` |
| this PR | `ok. 1 passed; 0 failed`, 603s |

The diagnosis is confirmed against the cluster rather than only against our own tree: on devnet today the same `verify_encoded_vaa_v1`, with the same guardian-set account and a 292-byte 3-signature VAA, succeeds. The deployed program accepts what the 2024 binary rejects.

### Not in this PR

- `test programs` needs a `PYTH_API_KEY` repo secret before it can exercise any oracle test; only an admin can add it.
- `external-programs/pyth-receiver.so` also diverges from the deployed program (709 424 vs 1 355 920 bytes). It does not break the tests today, so it is left alone here.
- The guardian-set freshness check added in [#423](https://github.com/gmsol-labs/gmx-solana/pull/423) has no counterpart for `external-programs/*.so`. Without one this rots again and is found by accident, which is how it went this time.


tests/anchor_test/setup.rs
```diff
@@ -243,7 +243,10 @@ impl Deployment {
             users: Users::new(&mut rng),
             extra_user_count: 0,
             rng,
-            hermes: Default::default(),
+            // Hermes rejects unauthenticated requests, so pick up `PYTH_API_KEY` when it is set.
+            // Falls back to the unauthenticated client when it is not, keeping the previous
+            // behaviour for contributors and CI runs without a key.
+            hermes: Hermes::from_default_envs().unwrap_or_default(),
```
angelo-75, line 249: suggestion: in fact, we can just error out here directly.

angelo-75 (approved):

PR #428, opened 2026-09-03
fix(sdk): skip the bundled close when a builder fee will be charged

crypto1gmt:
## What

`ExecuteOrderBuilder` merges a `CloseOrderV2` into the execution bundle by default (`close = true`).
For an order that gets charged a builder fee that reverts the whole execution: execution charges the
fee onto the order, and `CloseOrderV2::validate` rejects a non-zero `builder_fee_amount` with
`UnsettledBuilderFee`. The fee must first be settled through the separate, permissionless
`settle_builder_fee`.

This skips the bundled close when the order's checkpointed builder fee factor is non-zero, leaving
the order open for the keeper to settle and close afterwards. Orders that will not be charged are
unaffected.

## Why it matters

The CLI keeper runs `execute-order` with the bundled close on by default
(`crates/cli/src/commands/exchange/mod.rs:2205`), with no settle in its path, so before this a keeper
executing any fee-bearing order reverted with `UnsettledBuilderFee` unless the operator knew to pass
`--skip-close` and settle and close by hand.

## Change

- `ExecuteOrderHint` carries the order's `builder_fee_factor`, checkpointed by `set_builder_fee`
  before execution so it is observable at build time even though the fee is only charged during
  execution.
- The bundled close is gated on `hint.builder_fee_factor == 0`.
- A test covers the revoked case, which the suite did not reach before.
- `PositionCutBuilder` (liquidation/ADL) is untouched: `is_user_initiated_position` excludes those
  kinds, so `set_builder_fee` rejects them and they can never carry a fee.

## Why the factor and not the builder

`Order.builder` stays set through a revocation. Clearing a checkpoint is done by re-checkpointing
with a User Account that advertises `0`, and that writes the builder key like any other checkpoint.
The result is `builder = Some` with `builder_fee_factor = 0`, where execution charges nothing,
`builder_fee_amount` stays `0`, and the close succeeds. Gating on the builder's presence would skip a
close that works and leave the order open with no fee to settle, and the CLI has no
`settle_builder_fee` command to recover it. The on-chain code draws the same line, gating its
swap-type rule on `factor != 0` for this reason.

That state was already asserted in the suite before this PR: `order::set_builder_fee` checks
`revoked.builder == owner_user` and `revoked.builder_fee_factor == 0` together, so the old gate was
skipping the close for a state the tests already demonstrate is reachable.

## Known residual

The factor is read when the bundle is built, and `set_builder_fee` stays callable while the order is
pending. An owner attaching a fee between that read and the transaction landing still reverts the
execution, exactly as before. Callers that must not fail on it should pass `close(false)` and settle
and close separately.

## Test

`order::charges_builder_fee_on_execution` (added in #421, which merged with CI skipped so it never
ran) already covers the paying case: it relies on the default bundled close and then separately
exercises close-fails / settle / close, so the bundled close must not fire during execution.

`order::revoked_builder_fee_still_closes_in_bundle` is new here and covers the zero-factor case
argued above, which nothing exercised before: it checkpoints a zero-advertising account, executes
with the default bundled close, and asserts the order account is gone.

Oracle tests do not run on fork PRs (the `PYTH_API_KEY` secret is not exposed to them), so verified
locally, and the new test was run both ways to confirm it discriminates:

- Old gate (`hint.builder.is_none()`): `FAILED. 0 passed; 1 failed` in 99.95s, on `the bundled close
  must have closed an order that pays no fee`. The order was left open.
- New gate: `ok. 1 passed; 0 failed` in 98.63s.
- Whole module on the new gate: `anchor_test::order` gives `11 passed; 0 failed` in 592.78s,
  including `balanced_market_order` and `single_token_market_order` for the path that must still
  close.
- Before the fix, on `main` `ce245b9f`: `charges_builder_fee_on_execution` FAILED at `Instruction 4:
  custom program error: 0x17f2` (`UnsettledBuilderFee`).

### DIFF
--- a/crates/sdk/src/client/ops/exchange/order.rs
+++ b/crates/sdk/src/client/ops/exchange/order.rs
@@ -775,6 +775,15 @@ pub struct ExecuteOrderHint {
     store: Arc<Store>,
     market_token: Pubkey,
     position: Option<Pubkey>,
+    /// The order's checkpointed builder fee factor, zero when no fee is
+    /// payable. Written by `set_builder_fee` before execution, so it is
+    /// observable here even though the fee itself is only charged during
+    /// execution.
+    ///
+    /// The factor rather than the builder, because the two disagree: a
+    /// zero-factor checkpoint is how a builder fee is revoked, and it leaves
+    /// the builder set on an order that will never be charged.
+    builder_fee_factor: u128,
     owner: Pubkey,
     receiver: Pubkey,
     rent_receiver: Pubkey,
@@ -882,6 +891,10 @@ where
     }
 
     /// Set whether to close order after execution.
+    ///
+    /// Requesting the close is not a guarantee of one: it is skipped for an
+    /// order carrying a non-zero builder fee factor, whose fee has to be
+    /// settled by `settle_builder_fee` before a close can succeed.
     pub fn close(&mut self, close: bool) -> &mut Self {
         self.close = close;
         self
@@ -924,6 +937,7 @@ where
             store: store.clone(),
             market_token,
             position: optional_address(&params.position).copied(),
+            builder_fee_factor: order.builder_fee_factor,
             owner,
             receiver: order.header.receiver,
             rent_receiver,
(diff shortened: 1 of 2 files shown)


crypto1gmt:
Superseded by #430, re-opened from a branch in this repository so the CI jobs can reach the repo secrets (a `pull_request` run from a fork gets none, which is why `test programs` never really executed here). Same commit `e3df137f`, nothing rewritten. Closing this one to keep the review board clean.

PR #431, opened 2026-09-04
docs(store): document the builder fee invariants

crypto1gmt:
> Resubmitted from a branch in this repository so the CI jobs can reach the repo secrets. Supersedes #419, which ran from a fork and therefore never executed the oracle tests.

Documents the properties the builder fee mechanism upholds, as doc comments in the repo rather than as a document elsewhere, so they sit next to the code they are about and version with it.

## What is here

A module-level doc block on `states/order.rs` stating seven invariants: conservation, coverage, authorization, boundedness, delivery, isolation, liveness.

It states them and stops there. An earlier revision argued each one, and a closing section covered the two cases no transaction can reach; both are gone on review, the section listing what we believe to be true rather than proving it, since the proofs are inferred from the code.

Two existing comments already referred to "the charging invariant" and "a broken invariant" with nothing to point at. Both now name the **coverage** invariant and refer to the catalogue, folding "charging" into it so there is one name per property.

`UpdateOrderParams` gains a `CHECK:` annotation saying the field set is pinned, and a test pins it: the destructuring is exhaustive with no rest pattern, so a new field fails the build rather than silently widening what an update can reach. That matters because a fee is charged at the factor checkpointed on the order and taken from a bucket the decrease position swap type can empty, so neither value may become updatable without revisiting what a checkpoint means.

## Placement

Resolves CON-42.

The catalogue is a module doc on `states/order.rs`, and each enforcement site carries one line naming the invariant it upholds. The split was agreed on the issue.

The second commit adds that per-site layer across 14 sites, covering all seven invariants:

| Invariant | Sites |
|---|---|
| conservation | increase-path record + route, decrease-path record |
| coverage | `clamp_builder_fee_amount`, `charge_builder_fee_on_collateral_increment`, decrease-path record, the close guard |
| authorization | the factor-equality check in `set_builder_fee`, `UserHeader::set_builder_fee_factor` |
| boundedness | `compute_builder_fee_amount`, both cap enforcement points, `Factors::max_builder_fee_factor` |
| delivery | the settlement builder-equality check, the close guard |
| isolation | `compute_builder_fee_amount`, `estimate_builder_fee_for_collateral_withdrawal` |
| liveness | the feature-flag gate |

Convention: name the invariant, with the `crate::states::order` pointer on the first annotation in each file rather than repeated at all fourteen.

## Review rework

The last two commits apply the review (the second fixes an IDL graft, see the CHECK: thread): the per-invariant arguments, the closing "what cannot be exercised on chain" section, and a paragraph about charging having been wired by #421 are all removed, leaving the seven statements. The `UpdateOrderParams` doc is rewritten on the repo's `CHECK:` convention, and `gmsol_store.json` follows it, since that doc compiles into the IDL. The 14 per-site annotations are unchanged.

## Scope notes

- The IDL carries `docs` arrays, so any annotated item that appears in it is updated too. The delta is three entries across the two commits, `UpdateOrderParams`, `BuilderFeeSettled.settled_amount` and `Factors.max_builder_fee_factor`, grafted onto the checked-in file rather than copied wholesale; nothing else in the generated IDL differed.
- Nothing outside doc comments, one test and that IDL delta is touched. No behaviour changes.

### DIFF
--- a/programs/store/src/states/order.rs
+++ b/programs/store/src/states/order.rs
@@ -16,11 +16,11 @@
 //! order's final output token escrow; `settle_builder_fee` moves it to the
 //! builder and zeroes the record.
 //!
-//! **Charging is not yet wired.** Every execution call site currently passes a
-//! literal zero factor rather than reading the checkpoint, so the charging
-//! invariants below hold vacuously today. They describe the mechanism the
-//! checkpoint exists for, and become load-bearing when the call sites start
-//! reading it.
+//! **Charging is wired as of #421.** The execution call sites read the order's
+//! checkpointed factor instead of the literal zero they used to pass, so the
+//! charging invariants below are load-bearing rather than vacuous. An earlier
+//! revision of this document said the opposite and named that change as the
+//! moment they would become live; it landed, so they are.
 //!
 //! ## Conservation
 //!
--- a/programs/store/src/states/order.rs
+++ b/programs/store/src/states/order.rs
@@ -1,3 +1,148 @@
+//! Order accounts, the parameters that create and update them, and the
+//! invariants the builder fee mechanism upholds.
+//!
+//! # Builder fee invariants
+//!
+//! A builder fee is a share of an order's output paid to the interface that
+//! built the order. The mechanism is spread over several instructions and two
+//! execution paths, so the properties it guarantees are stated here once,
+//! together with the argument for each, rather than left implicit at the sites
+//! that enforce them. Comments elsewhere refer to these by name.
+//!
+//! The moving parts: a builder advertises a factor on its own User Account
+//! (`set_builder_fee_factor`); the order's owner checkpoints that builder and
+//! factor onto a pending order (`set_builder_fee`), which is the authorization;
+//! execution charges against the checkpoint and routes the amount into the
+//! order's final output token escrow; `settle_builder_fee` moves it to the
+//! builder and zeroes the record.
+//!
+//! **Charging is not yet wired.** Every execution call site currently passes a
+//! literal zero factor rather than reading the checkpoint, so the charging
+//! invariants below hold vacuously today. They describe the mechanism the
+//! checkpoint exists for, and become load-bearing when the call sites start
+//! reading it.
+//!
+//! ## Conservation
+//!
+//! The amount charged, the amount recorded on the order, the tokens in the
+//! escrow and the amounts on the events all agree.
+//!
+//! Charging is never a lone write. The increase path deducts the fee from the
+//! collateral increment and routes the same number into the final output token
+//! bucket in the same breath; `Order::record_builder_fee` carries that
+//! obligation as a `CHECK:` on itself, because settlement pays the recorded
+//! amount out of that escrow. The decrease path reaches the same state
+//! differently: it records the fee without netting it out of the output, so the
+//! full output lands in the escrow and the fee value is already sitting there.
+//! Recording accumulates with a checked add rather than overwriting, so several
+//! charges against one order sum. `BuilderFeeCharged` and `BuilderFeeSettled`
+//! each carry two amounts rather than one, so a divergence is observable on
+//! chain instead of inferred.
+//!
+//! ## Coverage
+//!
+//! From the moment a fee is recorded until it is settled, the escrow holds at
+//! least the recorded amount.
+//!
+//! On the increase path a fee larger than the collateral increment is refused
+//! outright (`BuilderFeeExceedsCollateral`) and the order is cancelled, since an
+//! increase has no partial outcome to fall back on. On the decrease path
+//! underpayment is tolerated instead, because a decrease may be closing an
+//! insolvent position, so the recorded amount is clamped to the output actually
+//! produced and therefore cannot exceed the escrow. In between, nothing can
+//! drain the escrow first: closing an order with a non-zero recorded amount is
+//! refused (`UnsettledBuilderFee`). Settlement clamps a second time as defence
+//! in depth, and under this invariant that clamp never bites.
+//!
+//! ## Authorization
+//!
+//! An order is never charged a fee its owner did not authorize, and never at a
+//! factor other than the one authorized.
+//!
+//! The factor used is the one checkpointed on the order, not a live read of the
+//! builder's account, so the builder's advertised rate may move afterwards
+//! without changing what an existing order owes. Writing that checkpoint takes
+//! the owner's signature and requires the builder's advertised factor to equal
+//! the factor the owner passed, exactly, which is what stops a builder raising
+//! its rate between the owner deciding and the transaction landing. A builder
+//! can only advertise on its own account, structurally: the User Account PDA
+//! seeds bind it to the signer, so doing otherwise is unconstructible rather
+//! than merely rejected. No update can reach the checkpoint or the swap type it
+//! depends on; see [`UpdateOrderParams`].
+//!
+//! ## Boundedness
+//!
+//! A charge never exceeds the payable fee, and no factor above the cap in force
+//! at checkpoint time is ever checkpointed.
+//!
+//! The store's cap is enforced twice, once when a builder advertises a rate and
+//! again when an owner checkpoints it, because the cap can be lowered in between
+//! and the checkpoint is the moment the rate becomes payable. Both treat a
+//! missing cap as zero, so a store that has never configured one denies every
+//! non-zero factor. The amount itself is bounded by what the order can pay, by
+//! the refusal on the increase path and the clamp on the decrease path. Note the
+//! computed amount rounds **up**, deliberately, so the fee is never
+//! under-collected; the bound on what is taken comes from those two mechanisms,
+//! not from the rounding.
+//!
+//! ## Delivery
+//!
+//! A settled fee reaches the checkpointed builder and nobody else, and an order
+//! cannot be closed while a fee is still owed.
+//!
+//! Settlement requires the passed User Account to equal the builder recorded on
+//! the order, and the destination is the associated token account of that
+//! builder and the final output token, so its owner and mint follow from the
+//! address derivation rather than from a check. The transfer is signed by the
+//! order PDA, so nothing outside the program can move the escrow. The close
+//! guard is the other half: an order with an unsettled fee must be settled
+//! first, since closing would otherwise sweep the fee to the receiver along with
+//! the regular output and destroy the record of who it belonged to.
+//!
+//! ## Isolation
+//!
+//! With no builder fee set, an order behaves as it did before the mechanism
+//! existed, and market-level accounting is untouched.
+//!
+//! Every fee behaviour sits behind a non-zero factor check, and the amount
+//! computation returns zero without even reading a price when the factor is
+//! zero. Only one pre-existing line changed rather than being added: the
+//! decrease path's collateral withdrawal now passes through the fee estimate,
+//! which returns its input unchanged at a zero factor, ahead of any arithmetic.
+//! Market accounting is separate and holds for a different reason: the withheld
+//! amount is deducted before the collateral reaches the position and is counted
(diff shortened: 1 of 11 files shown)


programs/store/src/states/order.rs
```diff
@@ -1,3 +1,148 @@
...
+//! factor onto a pending order (`set_builder_fee`), which is the authorization;
+//! execution charges against the checkpoint and routes the amount into the
+//! order's final output token escrow; `settle_builder_fee` moves it to the
+//! builder and zeroes the record.
+//!
+//! **Charging is wired as of #421.** The execution call sites read the order's
+//! checkpointed factor instead of the literal zero they used to pass, so the
+//! charging invariants below are load-bearing rather than vacuous. An earlier
+//! revision of this document said the opposite and named that change as the
+//! moment they would become live; it landed, so they are.
```
angelo-75: issue: This kind of information that includes temporary historical changes should not appear in the documentation. Please consider removing it.

programs/store/src/states/order.rs
```diff
@@ -1,3 +1,148 @@
...
+//! collateral increment and routes the same number into the final output token
+//! bucket in the same breath; `Order::record_builder_fee` carries that
+//! obligation as a `CHECK:` on itself, because settlement pays the recorded
+//! amount out of that escrow. The decrease path reaches the same state
+//! differently: it records the fee without netting it out of the output, so the
+//! full output lands in the escrow and the fee value is already sitting there.
+//! Recording accumulates with a checked add rather than overwriting, so several
+//! charges against one order sum. `BuilderFeeCharged` and `BuilderFeeSettled`
+//! each carry two amounts rather than one, so a divergence is observable on
+//! chain instead of inferred.
```
angelo-75: issue: I don't think these arguments should go here. Let the whole invariants section list only the invariants (the proofs are inferred from the code). Our goal for this section is to let the reader know what we believe to be true, not to prove that it is true.

programs/store/src/states/order.rs
```diff
@@ -1,3 +1,148 @@
...
+//!
+//! On the increase path a fee larger than the collateral increment is refused
+//! outright (`BuilderFeeExceedsCollateral`) and the order is cancelled, since an
+//! increase has no partial outcome to fall back on. On the decrease path
+//! underpayment is tolerated instead, because a decrease may be closing an
+//! insolvent position, so the recorded amount is clamped to the output actually
+//! produced and therefore cannot exceed the escrow. In between, nothing can
+//! drain the escrow first: closing an order with a non-zero recorded amount is
+//! refused (`UnsettledBuilderFee`). Settlement clamps a second time as defence
+//! in depth, and under this invariant that clamp never bites.
```
angelo-75: issue: Same as above.

programs/store/src/states/order.rs
```diff
@@ -1,3 +1,148 @@
...
+//! The factor used is the one checkpointed on the order, not a live read of the
+//! builder's account, so the builder's advertised rate may move afterwards
+//! without changing what an existing order owes. Writing that checkpoint takes
+//! the owner's signature and requires the builder's advertised factor to equal
+//! the factor the owner passed, exactly, which is what stops a builder raising
+//! its rate between the owner deciding and the transaction landing. A builder
+//! can only advertise on its own account, structurally: the User Account PDA
+//! seeds bind it to the signer, so doing otherwise is unconstructible rather
+//! than merely rejected. No update can reach the checkpoint or the swap type it
+//! depends on; see [`UpdateOrderParams`].
```
angelo-75: issue: Same as above.

programs/store/src/states/order.rs
```diff
@@ -1,3 +1,148 @@
...
+//!
+//! The store's cap is enforced twice, once when a builder advertises a rate and
+//! again when an owner checkpoints it, because the cap can be lowered in between
+//! and the checkpoint is the moment the rate becomes payable. Both treat a
+//! missing cap as zero, so a store that has never configured one denies every
+//! non-zero factor. The amount itself is bounded by what the order can pay, by
+//! the refusal on the increase path and the clamp on the decrease path. Note the
+//! computed amount rounds **up**, deliberately, so the fee is never
+//! under-collected; the bound on what is taken comes from those two mechanisms,
+//! not from the rounding.
```
angelo-75: issue: Same as above.

programs/store/src/states/order.rs
```diff
@@ -1,3 +1,148 @@
...
+//! the factor the owner passed, exactly, which is what stops a builder raising
+//! its rate between the owner deciding and the transaction landing. A builder
+//! can only advertise on its own account, structurally: the User Account PDA
+//! seeds bind it to the signer, so doing otherwise is unconstructible rather
+//! than merely rejected. No update can reach the checkpoint or the swap type it
+//! depends on; see [`UpdateOrderParams`].
+//!
+//! ## Boundedness
+//!
+//! A charge never exceeds the payable fee, and no factor above the cap in force
```
angelo-75: nit: "payable fee" is a bit ambiguous here.

programs/store/src/states/order.rs
```diff
@@ -1,3 +1,148 @@
...
+//! cannot be closed while a fee is still owed.
+//!
+//! Settlement requires the passed User Account to equal the builder recorded on
+//! the order, and the destination is the associated token account of that
+//! builder and the final output token, so its owner and mint follow from the
+//! address derivation rather than from a check. The transfer is signed by the
+//! order PDA, so nothing outside the program can move the escrow. The close
+//! guard is the other half: an order with an unsettled fee must be settled
+//! first, since closing would otherwise sweep the fee to the receiver along with
+//! the regular output and destroy the record of who it belonged to.
```
angelo-75: issue: Same as above.

programs/store/src/states/order.rs
```diff
@@ -1,3 +1,148 @@
...
+//! computation returns zero without even reading a price when the factor is
+//! zero. Only one pre-existing line changed rather than being added: the
+//! decrease path's collateral withdrawal now passes through the fee estimate,
+//! which returns its input unchanged at a zero factor, ahead of any arithmetic.
+//! Market accounting is separate and holds for a different reason: the withheld
+//! amount is deducted before the collateral reaches the position and is counted
+//! only into the transfer-out bucket, so it never enters a market's balances.
+//! The account layout supports this too, the checkpoint fields having been
+//! carved out of existing padding and reserved space, so an order written before
+//! the mechanism existed reads back as carrying no builder fee.
```
angelo-75: issue: Same as above.

programs/store/src/states/order.rs
```diff
@@ -1,3 +1,148 @@
...
+//! it is idempotent: a recorded amount of zero is an explicit no-op with no
+//! transfer and no CPI, which makes it safe to call in any state. It needs the
+//! builder's claim vault to already exist and deliberately creates nothing, so
+//! that precondition is discharged one instruction earlier, by `set_builder_fee`
+//! requiring the same account; a builder without one is refused at checkpoint
+//! time, where the owner simply does not get a builder attached, rather than at
+//! settlement time, where the order would already be stuck. The feature flag
+//! gates only the checkpoint, leaving settlement, claiming and execution
+//! ungated, so disabling the mechanism stops new orders taking on a fee without
+//! stranding a fee already owed.
```
angelo-75: issue: Same as above.

programs/store/src/states/order.rs
```diff
@@ -1,3 +1,148 @@
...
+//!
+//! ## What cannot be exercised on chain
+//!
+//! Two cases are argued rather than tested, because no transaction can reach
+//! them. There is no update path through which a forbidden swap type or a
+//! different factor could arrive, so the "reject it at update" case has no code
+//! to exercise; `UpdateOrderParams` pins that, and the compile fails if it
+//! stops being true. And no pending `Liquidation` or `AutoDeleveraging` order
+//! can exist to be checkpointed: order creation refuses both kinds, and the
+//! keeper builds a liquidation inside the transaction that executes it.
```
angelo-75: suggestion: It feels a bit out of place having this paragraph here, let's consider removing it.

programs/store/src/states/order.rs
```diff
@@ -30,6 +175,18 @@ pub use gmsol_utils::order::{OrderKind, OrderSide};
...
+///
+/// These are the only values an update may reach, and the builder fee rests on
+/// that twice over: a fee is charged at the factor checkpointed on the order,
+/// and it is taken from the final output token bucket, which
+/// [`DecreasePositionSwapType::CollateralToPnlToken`] empties. A factor
+/// reachable by an update would change the amount owed after the owner
+/// authorized it; a reachable swap type would let an order be checkpointed
+/// under a benign one and then flipped to one that pays nothing. Neither is
+/// reachable today, and `update_order_params_field_set_is_pinned` fails the
+/// build if that changes.
```
angelo-75: nit: We'd usually write this part more concisely, that is, using "CHECK" to annotate what should be checked and the reason why.

angelo-75 (changes requested):

programs/store/src/states/order.rs
crypto1gmt: Done in 9ccc35f8. Dropped the argument under all seven, including the six marked same as above, so the section is the statements and nothing else.

programs/store/src/states/order.rs
crypto1gmt: Removed in 9ccc35f8.

programs/store/src/states/order.rs
crypto1gmt: Reworded to "A charge never exceeds what the order can pay" in 9ccc35f8.

programs/store/src/states/order.rs
crypto1gmt: Removed in 9ccc35f8.

programs/store/src/states/order.rs
crypto1gmt: Rewritten as a CHECK: annotation in 9ccc35f8, 13 lines down to 8. It also moved `gmsol_store.json`, since that doc compiles into the IDL.

programs/store/src/states/order.rs
crypto1gmt: Coverage argument removed in 9ccc35f8.

programs/store/src/states/order.rs
crypto1gmt: Authorization argument removed in 9ccc35f8.

programs/store/src/states/order.rs
crypto1gmt: Boundedness argument removed in 9ccc35f8.

programs/store/src/states/order.rs
crypto1gmt: Delivery argument removed in 9ccc35f8.

programs/store/src/states/order.rs
crypto1gmt: Isolation argument removed in 9ccc35f8.

programs/store/src/states/order.rs
crypto1gmt: Liveness argument removed in 9ccc35f8.

programs/store/src/states/order.rs
crypto1gmt: Correction: that took a second commit, fc318316. Anchor strips `CHECK:` lines out of the generated IDL, so hand-grafting mine into `gmsol_store.json` made `check idls` red on 9ccc35f8. The annotation is now one self-contained sentence with the detail in a paragraph after it, so dropping the `CHECK:` line still leaves the IDL doc reading properly. Which is presumably why the rest of the repo keeps `CHECK:` to a single line.

angelo-75 (approved):

programs/store/src/states/order.rs
angelo-75: We need to find a way to handle this properly. Manually modifying the IDL is not a good idea.

PR #433, opened 2026-09-04
fix(store): drop the builder fee recorded by a discarded execution

crypto1gmt:
> Resubmitted from a branch in this repository so the CI jobs can reach the repo secrets. Supersedes #424, which ran from a fork and therefore never executed the oracle tests.

`Order` is not revertible. `perform_execution` takes it as a plain `&mut Order`, unlike `RevertiblePosition` and the markets, so anything written to it survives a soft failure. The builder fee was recorded part-way through execution, and when a later step fails with `cancel_on_execution_error` set, the `TransferOut` that was supposed to fund that fee is replaced by `Default::default()` while the record stays. The order is then cancelled, and every close path refuses it until `settle_builder_fee` has run against an escrow nothing was routed into.

Introduced by #421, which replaced four hardcoded zero fee factors with the order's checkpointed one; before that `record_builder_fee(0)` made the path unreachable.

### The fix

The execution no longer writes the fee onto the order until it has succeeded.

`execute_increase_position` and `execute_decrease_position` return the charged amount in `ExecutionFees`, alongside the `paid_fee_value` they already returned, and `perform_execution` calls `record_builder_fee` once, after the last fallible step, next to `unchecked_process_gt`. That is the shape `paid_fee_value` has always had: compute it during the attempt, apply it at the irreversible point.

The fee is still computed where it was, because both paths need it locally. Increase deducts it from the collateral before `position.increase()`, decrease clamps it against the realised output. Only the write to `Order` moved, and in a later round the event followed it, see "The event" below.

`record_builder_fee` runs just before `unchecked_process_gt` rather than after it. `unchecked_process_gt` mints into the store and the user part-way through, so an overflow raised after it had run would land in the soft-failure arm with that mint left behind.

An earlier revision of this PR snapshotted the amount before `perform_execution` and restored it in the soft-failure arm. The review on this PR pointed out that this trades a global property, that `perform_execution` leaves no state behind on failure, for a per-writer undo obligation, and that is the better reading: `paid_fee_value` carries exactly the same exposure to a fallible step being added later, so the builder fee was the only field defending itself differently. `restore_builder_fee_amount` goes away with it.

A hard failure needs no handling: the runtime reverts the account write.

### The event

Added after `mv-reyes` pointed out that the fix above splits a pair which used to agree. `BuilderFeeCharged` was still emitted where the fee is computed, several fallible steps above the record, and `emit_cpi` is a self-CPI: once it returns the event sits in the transaction's inner instructions, and the soft-failure arm commits that transaction rather than reverting it. A discarded execution therefore announced a charge the order never recorded. Reproduced before changing anything: `payable/paid = (13, 13)` against a recorded `0`.

The divergence is new in this PR rather than pre-existing. Before it the record was written next to the emit, so the two were wrong together; moving the record without the emit is what split them.

The charge now travels out in `ExecutionFees` as `Option<BuilderFeeCharge>` and is emitted inside the irreversible block. `Some` is set from the fee factor rather than from the amount, because the event's documented contract is that its presence separates "no builder attached" from "builder attached, nothing collectible", so a charge clamped to zero must still announce itself while recording nothing.

**The ordering inside that block is a trade, not a clean win.** The emit sits after `unchecked_process_gt` and immediately before `set_builder_fee_amount`. The write cannot fail, so a successful emit is always followed by its record and the two can no longer disagree. The residual is the reverse pair: an `Err` out of `emit_cpi` would leave GT counted for an execution that then cancels, which is the round 2 review point again. I took that over the alternative because placing the emit after the write reproduces the CON-59 shape, a record with nothing funding it, and because a self-CPI failing is not reachable through ordinary business outcomes the way the min-output rejection above the old emit site was. Narrowed rather than eliminated, and worth a second opinion.

### Tests

`soft_failed_execution_records_no_builder_fee` checkpoints a fee onto a `MarketDecrease` with an unreachable `min_output_amount`, which fails `validate_decrease_output_amounts` after the fee is charged and with `should_throw_error` left false, so the failure lands in the soft-failure arm. It asserts the order is `Cancelled`, that the recorded amount is unchanged, and that the order closes without settling anything. It now also asserts that no `BuilderFeeCharged` event was emitted, counted rather than looked up: the helper returns the number of matching events alongside the total read, so a run that read nothing at all fails on the total instead of passing on the absence.

`decrease_execution_records_builder_fee` is new, and is the half that was missing. The decrease path had no success-side coverage, and the assertion above, that a soft failure records zero, is equally satisfied by a decrease path that records nothing at all. A change to where the fee is written could therefore have stopped paying builders on every decrease order with the suite still green. The new test executes the same order shape with a reachable minimum and asserts the order is `Completed`, the recorded amount is non-zero, the `BuilderFeeCharged` event agrees with it, the close is refused until settlement, and the settled amount lands in the builder's claim vault.

The increase path already had both halves; `charges_builder_fee_on_execution` is its success side.

### Both ways

Each regression test was run against a deliberately broken build as well as the real one, since a test that has only ever passed proves nothing. Both breaks were injected into the code as it stands here, not into the earlier revision, and the two are different mistakes because the two tests guard opposite directions.

| test | injected break | broken build | real build |
| --- | --- | --- | --- |
| `soft_failed_execution_records_no_builder_fee` | record at the decrease charge site instead of deferring, i.e. the pre-fix behaviour | `FAILED`, `left: 13, right: 0` | `ok` |
| `decrease_execution_records_builder_fee` | drop the charged amount instead of returning it, so nothing is ever recorded | `FAILED`, order `Completed` carrying `0` | `ok` |

The first is the original bug reappearing: a discarded execution charging 13 against an order whose principal had been refunded. The second is the failure mode this restructure introduces the risk of, and the reason the test was added: builders silently paid nothing on every decrease order. `cargo clippy` does flag that second mistake (`unused variable`, `unused_mut`), but CI runs clippy without `-D warnings`, so the lint alone would not have stopped it.

One more thing worth flagging: the comment at `programs/store/src/instructions/builder_fee.rs:336` already asserts that "a failed execution charges a zero fee". That was false on `main` and is true again with this change.

### DIFF
--- a/programs/store/src/ops/order.rs
+++ b/programs/store/src/ops/order.rs
@@ -903,16 +903,9 @@ pub(crate) type ShouldSendTradeEvent = bool;
 
 /// The amounts an execution computes but must not write to the [`Order`] itself.
 ///
-/// The `Order` is the one piece of state `perform_execution` cannot roll back:
-/// the position, the swap markets and the virtual inventories are all revertible
-/// wrappers that only `commit()` at the end, while the order is passed down as a
-/// plain `&mut Order`, so anything written to it during a failed attempt outlives
-/// that attempt. Carrying these amounts out by value instead of writing them where
-/// they are computed keeps the guarantee that a failed execution leaves the order
-/// untouched, which is what the soft-failure arm relies on.
-///
-/// Both fields are applied together at the irreversible point, after the last
-/// fallible step of the execution has passed.
+/// CHECK: the `Order` is not revertible, unlike the position, the swap markets and
+/// the virtual inventories, so both fields are carried out by value and applied by
+/// the caller at the irreversible point rather than written where they are computed.
 struct ExecutionFees {
     /// Order and borrowing fee value paid, the basis for the GT reward.
     paid_fee_value: u128,
@@ -1077,7 +1070,7 @@ impl ExecuteOrderOperation<'_, '_> {
                     &mut market,
                     &mut swap_markets,
                     &mut transfer_out,
-                    &mut *self.order.load_mut()?,
+                    &*self.order.load()?,
                 )?;
                 market.commit();
                 false
@@ -1139,7 +1132,7 @@ impl ExecuteOrderOperation<'_, '_> {
                             &mut swap_markets,
                             &mut transfer_out,
                             &mut *event_loader.load_mut()?,
-                            &mut *self.order.load_mut()?,
+                            &*self.order.load()?,
                             builder_fee_factor,
                         )?;
                         (false, fees)
@@ -1151,7 +1144,7 @@ impl ExecuteOrderOperation<'_, '_> {
                         &mut swap_markets,
                         &mut transfer_out,
                         &mut *event_loader.load_mut()?,
-                        &mut *self.order.load_mut()?,
+                        &*self.order.load()?,
                         true,
                         Some(SecondaryOrderType::Liquidation),
                         builder_fee_factor,
@@ -1163,7 +1156,7 @@ impl ExecuteOrderOperation<'_, '_> {
                         &mut swap_markets,
                         &mut transfer_out,
                         &mut *event_loader.load_mut()?,
-                        &mut *self.order.load_mut()?,
+                        &*self.order.load()?,
                         true,
                         Some(SecondaryOrderType::AutoDeleveraging),
                         builder_fee_factor,
@@ -1177,7 +1170,7 @@ impl ExecuteOrderOperation<'_, '_> {
                         &mut swap_markets,
                         &mut transfer_out,
                         &mut *event_loader.load_mut()?,
-                        &mut *self.order.load_mut()?,
+                        &*self.order.load()?,
                         false,
                         None,
                         builder_fee_factor,
@@ -1190,20 +1183,9 @@ impl ExecuteOrderOperation<'_, '_> {
                     .load_mut()?
                     .update_with_transfer_out(&transfer_out)?;
 
-                // The irreversible point: every fallible step of the execution has
-                // passed, and what follows only commits the revertible wrappers. Both
-                // writes below land on accounts that cannot be rolled back, so this is
-                // the only place they are allowed to happen.
-                //
-                // The builder fee goes first because it is the only one of the two
-                // that can still fail without having written anything:
-                // `unchecked_process_gt` mints into the store and the user partway
-                // through, so an overflow raised after it had run would send the
-                // execution into the soft-failure arm with that mint left behind.
-                self.order
-                    .load_mut()?
-                    .record_builder_fee(fees.builder_fee_amount)?;
-
+                // The irreversible point. Both writes below land on accounts that
+                // cannot be rolled back, so they are held until every fallible step
+                // of the execution has passed.
                 if gt_minting_enabled {
                     self.order.load_mut()?.unchecked_process_gt(
                         &mut *self.store.load_mut()?,
@@ -1214,6 +1196,14 @@ impl ExecuteOrderOperation<'_, '_> {
                 } else {
                     msg!("[GT] GT minting is disabled for this market");
                 }
+
+                // Last, because it must be the last fallible call in this function:
+                // everything below only commits the revertible wrappers, and the
+                // fallible `handle_executed` runs outside the soft-failure arm, so a
+                // failure there reverts the transaction instead of discarding it.
+                self.order
+                    .load_mut()?
+                    .record_builder_fee(fees.builder_fee_amount)?;
 
                 position.commit();
                 msg!(
@@ -1566,7 +1556,7 @@ fn execute_swap(
     market: &mut RevertibleMarket<'_, '_>,
     swap_markets: &mut SwapMarkets<'_, '_>,
     transfer_out: &mut TransferOut,
-    order: &mut Order,
+    order: &Order,
 ) -> Result<()> {
     let swap_out_token = order
         .tokens
@@ -1645,7 +1635,7 @@ fn execute_increase_position(
     swap_markets: &mut SwapMarkets<'_, '_>,
     transfer_out: &mut TransferOut,
     event: &mut TradeData,
-    order: &mut Order,
+    order: &Order,
     builder_fee_factor: u128,
 ) -> Result<ExecutionFees> {
     // The builder fee is charged in the order's final output token, so for an increase order that
@@ -1751,8 +1741,6 @@ fn execute_increase_position(
     } else {
         (collateral_increment_amount, 0)
     };
-
-    let params = &order.params;
 
     // Increase position.
     let (long_amount, short_amount, paid_order_fee_value) = {
@@ -1869,7 +1857,7 @@ fn execute_decrease_position(
     swap_markets: &mut SwapMarkets<'_, '_>,
     transfer_out: &mut TransferOut,
     event: &mut TradeData,
-    order: &mut Order,
+    order: &Order,
     is_insolvent_close_allowed: bool,
     secondary_order_type: Option<SecondaryOrderType>,
     builder_fee_factor: u128,
--- a/programs/store/src/ops/order.rs
+++ b/programs/store/src/ops/order.rs
@@ -951,6 +951,19 @@ impl ExecuteOrderOperation<'_, '_> {
         let mut should_throw_error = false;
         let prices = self.prices()?;
         let discount = self.validate_and_get_order_fee_discount()?;
+
+        // The builder fee is recorded onto the order part-way through
+        // execution, and the order account is not revertible: unlike the
+        // position and the markets, it is passed down as a plain `&mut Order`,
+        // so a soft failure below keeps whatever was written to it while
+        // discarding the `TransferOut` that was supposed to fund the fee.
+        //
+        // Snapshotting here keys the record to the outcome rather than to an
+        // ordering of the fallible steps. Moving the record after the last one
+        // would fix the paths that exist today and silently rot the moment a
+        // new fallible step is added after it.
+        let builder_fee_amount_before = self.order.load()?.builder_fee_amount();
+
         let res = match self.perform_execution(&mut should_throw_error, prices, discount) {
             Ok((should_remove_position, mut transfer_out, should_send_trade_event)) => {
                 transfer_out.set_executed(true);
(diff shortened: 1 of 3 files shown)


angelo-75 (commented):
nit: The approach taken is indeed workable, but it makes us start to consider exactly which states need to be restored after an execution failure. Previously there was no such concern with perform_execution, since it generally guarantees that no state changes occur on failure. A better approach might still be to have the order finalize the fee update at the irreversible point, what do you think?

In fact, `paid_fee_value` is also an irreversible record on the Order, so I think we should refer to its solution (that is, record the change first, then apply the update at the irreversible point).

crypto1gmt:
Agreed, switched to that shape in c022fdb3.

`execute_increase_position` and `execute_decrease_position` now return the builder fee amount instead of writing it, and the order gets it once at the irreversible point, next to `unchecked_process_gt`. `restore_builder_fee_amount` is gone. There is no undo half any more because nothing is written during the attempt.

The fee is still computed where it was, since both paths need it locally: increase deducts it from the collateral before `position.increase()`, decrease clamps it against the realised output. Only the write to `Order` moved.

One ordering detail: `record_builder_fee` goes just before `unchecked_process_gt`, not after. `unchecked_process_gt` mints into the store and the user part-way through, so an overflow raised after it had run would drop into the soft-failure arm with that mint left behind. It also sits after `write_to_event` and `update_with_transfer_out`, which is what the old description said the "last fallible step" approach would miss.

On my old comment claiming a later record "would silently rot the moment a new fallible step is added after it", I withdraw it. `paid_fee_value` has exactly the same exposure and the codebase already leans on that ordering everywhere, so having this one field defend itself differently was worse than following the convention. Your version is also strictly stronger than mine: with the snapshot `perform_execution` still dirtied the order and cleaned up afterwards, now it never writes until it has succeeded.

Also added `decrease_execution_records_builder_fee`. The decrease path had no success-side test, only the soft-failure one asserting zero, and a path that records nothing at all satisfies that assertion too. So this restructure could have stopped paying builders on every decrease order with the suite still green. Ran it both ways: with the propagation deliberately broken it fails on its own assertion (order `Completed`, recorded `0`), and it passes on the real thing. Increase already had its pair in `charges_builder_fee_on_execution`. `cargo clippy` does flag that particular break, but CI runs clippy without `-D warnings`, so the lint alone would not have caught it.

One separate thing your framing surfaced, not touching it here: `BuilderFeeCharged` is emitted where the fee is computed, before the steps that can still fail, so a soft failure should leave an event behind for a charge that never gets recorded. I have not confirmed that the emitted event actually survives the swallowed error, so treat that half as unverified. Either way it is not specific to the builder fee, `PositionIncreased` is emitted before `validate_market_balances` in the same function. Happy to check it properly and file it separately.

programs/store/src/ops/order.rs
```diff
@@ -901,6 +901,18 @@ pub(crate) struct ExecuteOrderOperation<'a, 'info> {
 pub(crate) type RemovePosition = bool;
 pub(crate) type ShouldSendTradeEvent = bool;
 
+/// The amounts an execution computes but must not write to the [`Order`] itself.
+///
+/// CHECK: the `Order` is not revertible, unlike the position, the swap markets and
+/// the virtual inventories, so both fields are carried out by value and applied by
+/// the caller at the irreversible point rather than written where they are computed.
+struct ExecutionFees {
```
angelo-75, line 909: nit: There might be a better name, but it's private so it's not a big deal.

programs/store/src/ops/order.rs
```diff
@@ -1171,17 +1183,28 @@ impl ExecuteOrderOperation<'_, '_> {
...
                     msg!("[GT] GT minting is disabled for this market");
                 }
 
+                // Last, because it must be the last fallible call in this function:
+                // everything below only commits the revertible wrappers, and the
+                // fallible `handle_executed` runs outside the soft-failure arm, so a
+                // failure there reverts the transaction instead of discarding it.
+                self.order
+                    .load_mut()?
+                    .record_builder_fee(fees.builder_fee_amount)?;
```
angelo-75: issue: The second write can still fail, causing GT to be over-counted while the execution actually failed. So I think here we should:

1. Have only a single borrow of the Order
2. Make record_builder_fee non-failing (e.g. use replace instead of add, reading first, computing the result with no error, and only then writing)

suggestion: Perhaps we could first check whether builder_fee_amount is 0 before considering the write. This would avoid unnecessary side effects.

angelo-75 (changes requested):

programs/store/src/ops/order.rs
crypto1gmt: Fixed in `4fac770e`. You were right that the previous commit moved the hazard rather than removing it.

`record_builder_fee` is split in two. `builder_fee_amount_after` does the `checked_add` and runs before the irreversible point, where a failure is still discarded cleanly; `set_builder_fee_amount` is a plain store with no `Result`. Nothing fallible is left after `unchecked_process_gt`.

Both writes now sit under a single `self.order.load_mut()?` held for the block, and the write is skipped entirely when the charged amount is `0`, so a fee-less execution never touches the field.

On the reasoning: I had talked myself out of this because the field is `0` when execution starts, so the add is always `0 + amount` and cannot overflow. That is still true, but it makes the guarantee depend on reachability instead of structure, which is the thing this PR is trying to get away from.

programs/store/src/ops/order.rs
crypto1gmt: Left as is. The alternatives I considered, `DeferredOrderWrites` and `PendingOrderUpdates`, describe when the struct is applied rather than what it holds, and both fields really are fees the execution computed. Happy to rename if you have one in mind.

mv-reyes:
Nice fix. Moving execute_* to &Order so the no-writes-on-failure property is enforced by the compiler is a good pattern.

One thing I think this change introduces: BuilderFeeCharged is still emitted where the fee is computed, but the record only happens in the irreversible block now. On a soft failure after the emit, the transaction commits with the event in it while the order records nothing and nothing ever settles.

Both paths have it. Decrease emits at ops/order.rs:2077 and the failing validate_decrease_output_amounts is right after at :2086. Increase has a wider window, the emit at :1744 comes before position.increase, the funding transfers, and validate_market_balances.

This is observable because the event goes out as an inner instruction via emit_cpi_with_space, so it sits in committed tx metadata for any indexer. The doc on the event says indexers distinguish "no builder attached" from "builder attached, nothing collectible" by whether this event is present at all, and last_builder_fee_charged in the test suite treats presence as proof of charge. A soft-failed execution now produces a charge event with no charge behind it.

You can see it with one line in soft_failed_execution_records_no_builder_fee: call last_builder_fee_charged after the cancelled execution and it returns the 13 that was never recorded.

Not a money bug, BuilderFeeSettled still only fires when value actually moves. But anything summing Charged events, or reconciling Charged minus Settled, over-counts from this version on.

Suggested fix: move the emit down next to set_builder_fee_amount so the event only exists when the record does. If you would rather the event keep reflecting the attempted charge, then the doc and an event assertion in the soft-failure test should pin that behavior instead.

crypto1gmt:
Confirmed, and thanks for the precise repro. I put your one-liner into `soft_failed_execution_records_no_builder_fee` at `4fac770e`, asserting the event is *absent* so that the assertion firing is the finding rather than something to argue about:

```
panicked at tests/anchor_test/order.rs:2088:
PROBE: BuilderFeeCharged survived the soft failure, payable/paid = Ok((13, 13))
test result: FAILED. 0 passed; 1 failed; 53 filtered out; finished in 96.28s
```

13 on the nose. Both windows are as you described: decrease emits at `ops/order.rs:2077` with `validate_decrease_output_amounts` nine lines later, increase emits at `:1744` with `position.increase`, the funding transfers and `validate_market_balances` all still ahead of it.

Worth saying plainly that the divergence is new in this PR, not pre-existing. Before it the record was written next to the emit, so the two agreed and were wrong together; moving the record without the emit is what split them. That is your reading and it is right.

Direction, unless the lead wants otherwise: move the emit into the irreversible block next to `set_builder_fee_amount`, carrying the token and both amounts out through the same `ExecutionFees` struct the recorded amount already travels in. Emit goes *after* the record, so the worst remaining case is an event missing for a record that exists, which under-counts instead of over-counting.

One thing has to be settled first, and it cuts against an earlier review request on this same PR. The doc on `BuilderFeeCharged` requires the event even when `paid_amount` is 0, fully clamped by a shortfall, so indexers can tell "no builder attached" from "builder attached, nothing collectible". The irreversible block currently gates on `builder_fee_amount == 0` and skips the write. Hanging the emit on that gate would drop the event in exactly the case the doc calls out, so it has to key on the fee factor instead.

Separately, #431 is on main with a Conservation invariant saying the charged amount, the recorded amount, the escrow and the event amounts all agree. On the soft-failure path that is false today, so the wording moves with the fix.

programs/store/src/ops/order.rs
```diff
@@ -1171,15 +1207,68 @@ impl ExecuteOrderOperation<'_, '_> {
...
+                    // outcomes the way the min-output rejection below the old
+                    // emit site was.
+                    if let Some(charge) = &fees.builder_fee {
+                        position.event_emitter().emit_cpi(&BuilderFeeCharged::new(
+                            order.header().store(),
+                            &position.market().market_meta().market_token_mint,
+                            &charge.token,
+                            charge.payable_amount,
+                            charge.paid_amount.into(),
+                        )?)?;
```
angelo-75: issue (non-blocking): Although it's actually unlikely to happen, to let the reader quickly verify this, the two `?`s here should still be considered for changing to `.expect`.

angelo-75 (dismissed):

programs/store/src/ops/order.rs
crypto1gmt: Done in `6f1ba610`. Both `?` are `.expect` now, and the comment above them no longer claims a residual, because there is not one any more: a `?` there would have been swallowed by the soft-failure arm and committed, leaving the GT mint standing for an execution that then cancels, while a panic aborts the transaction instead. So the ordering trade I raised in chat is closed rather than narrowed. Thanks, that is a better answer than any of the three orderings I had been choosing between.

The two `?` still in the block, `load_mut` and `unchecked_process_gt`, both sit before any write, which is the shape your own comment on that block describes, so I left them.

angelo-75 (approved):

PR #435, opened 2026-09-07
fix(store): make settle_builder_fee's transferring accounts optional

crypto1gmt:
`settle_builder_fee` documents itself as a universal no-op, "safe to call in any order state", but neither the accounts struct nor the SDK let that call be constructed for an order whose final output token escrow was never initialized. `final_output_token` and `escrow` were required while `builder_user` and `claim_vault` were optional, although all four are used only on the transferring path. Orders without an escrow are a supported case, covered by `increase_order_without_escrow_leaves_final_output_token_uninitialized`, so the no-op was unreachable for exactly the orders that need it.

That matters beyond tidiness: `CloseOrderV2::validate` requires a zero recorded fee on every close path, so settlement sits on the critical path to closing an order and reclaiming rent.

### The predicate

All four accounts are optional and required when the order's **recorded builder fee amount** is non-zero, not when a builder is present. A non-zero fee already implies the escrow exists, because `set_builder_fee` refuses to checkpoint a builder onto an order whose final output token account is uninitialized (`BuilderFeeFinalOutputTokenEscrowNotInitialized`), so keying all four on that one value is sufficient and internally consistent.

Requiring the builder instead was considered and rejected: it would make the instruction unsafe to call on an order that never had one, contradicting the permissionless and idempotent property the handler's own doc comment promises.

### The program

`final_output_token` and `escrow` become `Option<...>`, and all four are unwrapped in the handler after the zero check, so the no-op stays reachable for an order that has none of them. **Nothing moves from a declarative constraint into the handler**: the `associated_token::mint` and `associated_token::authority` constraints stay where they are, and Anchor applies them when the account is present. `close_order_v2` already carries four optional escrow accounts in exactly that shape, three of them referencing mints that are themselves optional.

The entrypoint's `# Errors` list now names all four accounts as required when the recorded fee is non-zero. It previously named only `builder_user` and `claim_vault`, which stopped being the whole story the moment the other two became optional, and that list compiles into the IDL. The mint's own unwrap uses `TokenMintNotProvided` rather than `TokenAccountNotProvided`, matching `execute_order`, which unwraps a vault and a mint of this same type on adjacent lines.

### The SDK

The hint's `final_output_token` and `escrow` become `Option<Pubkey>`, and building no longer fails when the order has neither. The old error, `order has no final output token`, was raised for the mint and for the escrow alike; it is now three separate messages, each reachable only with a non-zero recorded fee, so a caller-supplied hint missing one of them says which. A hint derived from the order cannot miss just one: `TokenAndAccount` records the mint and the account together or records neither.

Building the instruction goes through `fix_optional_account_metas`, as every op with optional accounts must: Anchor emits `AccountMeta::new_readonly(crate::ID, false)` for a `None` account, fixed at codegen time, while the program decides `None` against the runtime program id, so the two diverge for a client built with a `store_program_id` override.

### The IDL

Regenerated in the same change, since account optionality is expressed there as `optional` per account. The delta is confined to `settle_builder_fee`, asserted structurally rather than by eye, and `settle_builder_fee` appears in no other program's IDL. It carries the corrected `# Errors` text as well, since entrypoint doc comments compile into the IDL.

### The changelog

No entry. Both `settle_builder_fee` and `SettleBuilderFeeHint` arrived in #408, which is also unreleased, so there is nothing here a 0.10 consumer would see change. The two entries this PR originally carried were removed in review.

### Tests

`settle_builder_fee_is_a_no_op_without_escrow` is the coverage this ticket exists for and was **previously unconstructible**: the SDK refused before building anything. `settle_builder_fee_is_a_no_op_with_escrow_and_zero_fee` is kept separate on purpose, because the two fail differently if the predicate is ever changed from the recorded amount to the presence of a builder: one would still build and start touching accounts, the other would not build at all.

Both take the exclusive locked user rather than `DEFAULT_USER`. Every test in that module mints into and spends from the shared balance with no lock over it, and adding two more consumers of it made the known `0x1` InsufficientFunds contention fire on order creation.

### One AC line without a test, deliberately

The AC says a non-zero fee meeting a missing account must fail distinguishably. That path is **unreachable through the SDK**, which now refuses first with its own message, so the program's `TokenAccountNotProvided` guard remains as defence in depth but has no test; covering it would mean building the instruction by hand, which nothing in that test file does today. Raised in chat and accepted rather than quietly closed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Breaking Changes
 
 - programs(store): Creating an increase order now requires its final output token to be the position's collateral token, and executing one whose final output token was recorded at creation revalidates the same thing. Creating an order with a different final output token used to succeed and silently ignore the value; it now reverts with `TokenMintMismatched`. Existing orders with an uninitialized final output token are unaffected and keep executing.
+- sdk(sdk): Changed `SettleBuilderFeeHint`'s `final_output_token` and `escrow` fields from `Pubkey` to `Option<Pubkey>`. Both are `None` for an order whose final output token escrow was never initialized, which is now a hint the SDK builds rather than an error it raises. Code constructing the hint by hand must wrap the two values.
 
 ### Added
 
@@ -28,6 +29,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - programs(store): An increase order now records its final output token at creation when the escrow is provided, which is what makes it eligible for a builder fee later.
 - sdk(solana-utils): Kept the two-argument `Bundle::send_all_with_opts` as a deprecated compatibility wrapper around the detailed API. It still returns the compressed success-signature list, and when multiple transactions fail it returns the **last** real send error (matching prior overwrite semantics; `SendAborted` placeholders are ignored).
+- programs(store): Made `settle_builder_fee`'s `final_output_token` and `escrow` accounts optional, joining `builder_user` and `claim_vault`. All four are required exactly when the order's recorded builder fee amount is non-zero, so the instruction's documented no-op is now reachable for an order whose final output token escrow was never initialized. Such orders are a supported state, and settlement sits on the critical path to closing an order because `CloseOrderV2` rejects a non-zero recorded fee.
 
 ## [0.10.0] - 2026-08-12
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Breaking Changes
 
 - programs(store): Creating an increase order now requires its final output token to be the position's collateral token, and executing one whose final output token was recorded at creation revalidates the same thing. Creating an order with a different final output token used to succeed and silently ignore the value; it now reverts with `TokenMintMismatched`. Existing orders with an uninitialized final output token are unaffected and keep executing.
-- sdk(sdk): Changed `SettleBuilderFeeHint`'s `final_output_token` and `escrow` fields from `Pubkey` to `Option<Pubkey>`. Both are `None` for an order whose final output token escrow was never initialized, which is now a hint the SDK builds rather than an error it raises. Code constructing the hint by hand must wrap the two values.
 
 ### Added
 
@@ -29,7 +28,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - programs(store): An increase order now records its final output token at creation when the escrow is provided, which is what makes it eligible for a builder fee later.
 - sdk(solana-utils): Kept the two-argument `Bundle::send_all_with_opts` as a deprecated compatibility wrapper around the detailed API. It still returns the compressed success-signature list, and when multiple transactions fail it returns the **last** real send error (matching prior overwrite semantics; `SendAborted` placeholders are ignored).
-- programs(store): Made `settle_builder_fee`'s `final_output_token` and `escrow` accounts optional, joining `builder_user` and `claim_vault`. All four are required exactly when the order's recorded builder fee amount is non-zero, so the instruction's documented no-op is now reachable for an order whose final output token escrow was never initialized. Such orders are a supported state, and settlement sits on the critical path to closing an order because `CloseOrderV2` rejects a non-zero recorded fee.
 
 ## [0.10.0] - 2026-08-12
 
--- a/crates/sdk/src/client/ops/builder_fee.rs
+++ b/crates/sdk/src/client/ops/builder_fee.rs
@@ -1,14 +1,20 @@
 use std::{future::Future, ops::Deref};
 
 use anchor_spl::associated_token::get_associated_token_address_with_program_id;
-use gmsol_programs::gmsol_store::client::{accounts, args};
+use gmsol_programs::gmsol_store::{
+    client::{accounts, args},
+    ID,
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Breaking Changes
 
 - programs(store): Creating an increase order now requires its final output token to be the position's collateral token, and executing one whose final output token was recorded at creation revalidates the same thing. Creating an order with a different final output token used to succeed and silently ignore the value; it now reverts with `TokenMintMismatched`. Existing orders with an uninitialized final output token are unaffected and keep executing.
+- sdk(sdk): Changed `SettleBuilderFeeHint`'s `final_output_token` and `escrow` fields from `Pubkey` to `Option<Pubkey>`. Both are `None` for an order whose final output token escrow was never initialized, which is now a hint the SDK builds rather than an error it raises. Code constructing the hint by hand must wrap the two values.
```
angelo-75: issue (non-blocking): `SettleBuilderFeeHint` did not exist in the previous version (0.10), so it does not count as a change.

CHANGELOG.md
```diff
@@ -28,6 +29,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - programs(store): An increase order now records its final output token at creation when the escrow is provided, which is what makes it eligible for a builder fee later.
 - sdk(solana-utils): Kept the two-argument `Bundle::send_all_with_opts` as a deprecated compatibility wrapper around the detailed API. It still returns the compressed success-signature list, and when multiple transactions fail it returns the **last** real send error (matching prior overwrite semantics; `SendAborted` placeholders are ignored).
+- programs(store): Made `settle_builder_fee`'s `final_output_token` and `escrow` accounts optional, joining `builder_user` and `claim_vault`. All four are required exactly when the order's recorded builder fee amount is non-zero, so the instruction's documented no-op is now reachable for an order whose final output token escrow was never initialized. Such orders are a supported state, and settlement sits on the critical path to closing an order because `CloseOrderV2` rejects a non-zero recorded fee.
```
angelo-75: issue (non-blocking): `settle_builder_fee` did not exist in the previous version (0.10), so it does not count as a change.

tests/anchor_test/order.rs
```diff
@@ -2348,3 +2353,168 @@ async fn decrease_execution_records_builder_fee() -> eyre::Result<()> {
 
     Ok(())
 }
+
+/// `settle_builder_fee` is buildable and succeeds as a no-op on an order whose final output token
+/// escrow was never initialized.
+///
+/// This is the case CON-46 exists for, and it was previously **unconstructible**: the accounts
```
angelo-75: issue: must not contain local identifiers.

crates/sdk/src/client/ops/builder_fee.rs
```diff
@@ -85,54 +93,71 @@ impl<C: Deref<Target = impl Signer> + Clone> BuilderFeeOps<C> for crate::Client<
...
-                &hint.final_output_token,
-                &anchor_spl::token::ID,
-            );
-            (Some(builder), Some(claim_vault))
-        };
+            };
 
         let rpc = self
             .store_transaction()
             .anchor_accounts(accounts::SettleBuilderFee {
```
angelo-75: issue: this should use `fix_optional_account_metas` because optional accounts are involved.

tests/anchor_test/order.rs
```diff
@@ -2348,3 +2353,168 @@ async fn decrease_execution_records_builder_fee() -> eyre::Result<()> {
...
+    assert!(
+        before.tokens.final_output_token.token_and_account().is_none(),
+        "the fixture must produce an order with no final output token, otherwise this test is vacuous"
+    );
+    assert_eq!(
+        before.builder_fee_amount, 0,
+        "an order that never had a builder must record no fee"
+    );
+
+    // The assertion is that this builds at all: before CON-46 the SDK errored here.
```
angelo-75: issue: must not contain local identifiers.

angelo-75 (changes requested):

crates/sdk/src/client/ops/builder_fee.rs
crypto1gmt: Done.

Confirmed the mechanism while fixing it, in case it's useful: anchor-syn emits
`AccountMeta::new_readonly(crate::ID, false)` for a `None` optional account
(`codegen/accounts/to_account_metas.rs:27`), where `crate::ID` is fixed at codegen time, while
the program decides an account is `None` by comparing it against the runtime `__program_id`
(`codegen/accounts/try_accounts.rs:50`). So the two only diverge for a client built with a
`store_program_id` override, which also means the suite can't catch a regression here: the tests
run under the declared id, `fix_optional_account_metas` takes its `original == current` early
return, and everything passes either way.

CHANGELOG.md
crypto1gmt: Right, dropped it. `SettleBuilderFeeHint` came in with #408, which is also unreleased, so there's nothing here a 0.10 user would see change.

CHANGELOG.md
crypto1gmt: Dropped this one too, same reason.

tests/anchor_test/order.rs
crypto1gmt: Removed.

tests/anchor_test/order.rs
crypto1gmt: Removed.

angelo-75 (approved):

PR #436, opened 2026-09-09
docs(store): record the frozen claim vault liveness exception

crypto1gmt:
The builder fee invariants say settlement cannot be blocked and an order is always eventually closable. Neither holds against the fee token's freeze authority. Settlement must transfer into the builder's claim vault, SPL Token rejects a transfer into a frozen account before it reads the amount, and the close guard lifts only once the fee is settled. The vault is the ATA of a single mint owned by the builder's User Account, so freezing it leaves unclosable every executed-but-unclosed order that owes that builder a fee in that mint, for as long as the freeze stands. Orders owing the same builder in another mint are unaffected.

This records that as a known exception. **No behavioural change.** The remedy is a governance waiver of a fee owed to a given builder, and it is deliberately not in this PR.

The report cites three doc comments as asserting the opposite of its findings, its points 5, 7 and 8. All three are corrected here.

What changes:

- `states/order.rs`: the Liveness invariant is qualified to what the protocol controls, and a `Known exception` section states the failure, who bears it (the order owners whose escrows are held, not the frozen builder, and every escrow on the order rather than the fee alone; for a decrease order that is the whole position payout, which may split between the final output token escrow and the pnl token's), and that nothing in the protocol lifts it short of a program upgrade, only the freeze authority thawing the vault.
- `settle_builder_fee`'s entrypoint doc in `lib.rs`: "no specific address can block it" was the exact claim this PR retracts. It now says no authority *within the protocol* can block settlement, and names the freeze authority that can.
- `SetBuilderFee::claim_vault` (report point 5): "so that settlement cannot fail later" becomes "for want of a destination", followed by the point that existence is necessary and not sufficient.
- `SetBuilderFee::builder` (report point 8): the doc promised that "an owner is never bound to a builder for the life of the order". That is false over the interval that matters, since this instruction requires the order to still be pending while a fee is recorded only during execution, which leaves the pending state in the same instruction. It now says the escape closes at execution, and that detaching a builder is not a route around the exception.
- The clamp comment in `settle_builder_fee`: scoped to a shortfall in the escrow, and it now says what it does not cover.
- The feature-flag comment in `SetBuilderFee::invoke` (report point 7): it said disabling the mechanism strands nothing already owed, which is true, without saying that the flag is no lever in the other direction either. Disabling the feature cannot release an order whose settlement is already blocked.
- Two comments used "freeze" to mean seizing up under the feature flag, which now sits next to a section using it for an SPL freeze. Reworded to "strand".
- The acceptance rationale now says outright that a claim vault belonging to a PDA is a statement about likelihood and not capability: a freeze applies to the token account whoever its authority is.
- `gmsol_store.json` regenerated. Three docs reach it, the `claim_vault` and `builder` account docs on `set_builder_fee` plus the `settle_builder_fee` instruction doc; module-level docs reach no IDL, so the invariants text itself produces no delta.

Rebased onto main after #435 merged. The only conflict was in `builder_fee.rs`, not `gmsol_store.json`: #435 unwraps the escrow into a local binding, so the line under the clamp comment is theirs and the comment is this PR's. The IDL merged clean and keeps the `optional` flags #435 added to `settle_builder_fee`.


programs/store/src/states/order.rs
```diff
@@ -48,8 +48,34 @@
...
+//! Liveness does not hold against the fee token's freeze authority. Settlement
+//! must transfer into the builder's claim vault, SPL Token rejects a transfer
+//! into a frozen account before it reads the amount, and the close guard under
+//! `Delivery` lifts only once the fee is settled. Freezing one builder's claim
+//! vault therefore leaves every executed-but-unclosed order checkpointed to
+//! that builder unclosable for as long as the freeze stands. Nothing in the
+//! protocol can lift it at any privilege level short of a program upgrade; only
+//! the freeze authority itself can, by thawing the vault.
+//!
+//! The loss falls on the order owners whose escrows are held, not on the frozen
```
angelo-75: issue: shouldn't describe it as a loss here, since the funds are held rather than lost and a thaw returns them. Something like "The impact falls on the order owners whose escrows are held" keeps it consistent with the wording below.

programs/store/src/states/order.rs
```diff
@@ -48,8 +48,34 @@
...
+//! owed.
+//!
+//! ### Known exception: a frozen claim vault
+//!
+//! Liveness does not hold against the fee token's freeze authority. Settlement
+//! must transfer into the builder's claim vault, SPL Token rejects a transfer
+//! into a frozen account before it reads the amount, and the close guard under
+//! `Delivery` lifts only once the fee is settled. Freezing one builder's claim
+//! vault therefore leaves every executed-but-unclosed order checkpointed to
+//! that builder unclosable for as long as the freeze stands. Nothing in the
```
angelo-75: issue: the wording here is too broad. The claim vault is the ATA of that specific mint owned by the builder_user account (PDA), so a freeze only affects orders whose fee is owed in that mint, not every order checkpointed to that builder_user.

angelo-75 (changes requested):

programs/store/src/states/order.rs
crypto1gmt: Done, took your wording. The clause right after it already said "held", so "loss" was just wrong there. 8da978ae

programs/store/src/states/order.rs
crypto1gmt: Fixed in 8da978ae. It now says the vault is the ATA of one mint owned by the `builder_user`, so a freeze only reaches orders owing that builder a fee in that mint, and it spells out that orders owing the same builder in another mint are unaffected. The PR description had the same overbroad sentence, fixed that too.

angelo-75 (approved):

PR #439, opened 2026-09-11
fix(sdk): make liquidation price agree with the contract's liquidation condition

crypto1gmt:
Makes `PositionStatus.liquidation_price` agree with the price at which `check_liquidatable(&prices, true, true)` actually flips. Two independent defects, plus a sweep that measures the result against the contract's own check.

### 1. The threshold came from the wrong collateral factor

The SDK built it from `min_collateral_factor`, but a liquidation order is checked against `min_collateral_factor_for_liquidation` (`crates/model/src/position.rs:526`). That accessor falls back to the plain factor only when the market config leaves the field at 0, and no mainnet market does: all 101 set it, 89 of them to exactly half. So the reported price was wrong for every position on almost every market, independently of which token is posted as collateral. It read higher than reality for a long, warning earlier than the contract acts.

The market-closed variant needs no extra code: `position_params()` already resolves `min_collateral_factor_for_liquidation(is_closed())`.

### 2. Same-token collateral was solved against a price-pinned collateral value

When the collateral token *is* the index token, `collateral_value` and the pending funding fee are both `amount × collateral_token_price`, so they move with the very price being solved for and cannot be held at spot. Everything else in the expression is a plain USD amount (the borrowing fee and the close order fee are both `apply_factor(..)` over a USD size, and the price impact comes off pool balances), so the boundary stays linear in `P` and the correction lands entirely in the denominator, plus a non-negative `surcharge = borrowing + order_fee - impact` in the numerator. Positions collateralised in a different token take the same path as before.

### Measured against the real check, before and after

`sweep_reported_vs_real_boundary` binary-searches `check_liquidatable(&prices, true, true)` for the price at which it flips, then compares that boundary with what `status()` reports, over long/short × same-token/different-token × zero/50bps oracle spread, with non-zero fees throughout. Error as a fraction of the boundary:

| shape | before | after |
|---|---|---|
| long, same token, 0bps | 0.934% | **0.000%** |
| long, same token, 50bps | 1.155% | **0.250%** |
| short, same token, 0bps | 0.946% | **0.000%** |
| short, same token, 50bps | 0.666% | **0.306%** |
| different token, 0bps | 0.000% | 0.000% |
| different token, 50bps | 0.250% | 0.250% |

The different-token rows are the control: identical before and after, since that path is untouched.

The residual under a spread is not introduced here. The formula solves for a single price while the model picks `.min` for collateral and `.min`/`.max` for pnl by side, so roughly a half-spread of error is inherent and shows up on positions this change does not touch as well. At a zero spread the reported price is the boundary exactly.

### Tests

- two factors differing: the reported price is built from the liquidation factor, with an `assert_ne!` guard against the value the old code produced
- the same case pinned to numbers, 18.10 against the 18.20 the old code gave, on the config of a real mainnet market
- the field left at 0: the fallback still works
- same-token collateral: the reported price must be the real boundary
- same-token collateral with a non-zero order fee, which is the case the first version of commit 2 got wrong
- the sweep above, asserting exactness at zero spread, a half-spread ceiling otherwise, and that all eight shapes were reached rather than skipped
- a closed market: the threshold comes from `market_closed_min_collateral_factor_for_liquidation`, with an `assert_ne!` guard against the open-market factor
- a binding `min_collateral_value` floor on the same-token path, which every other fixture here leaves at 0, asserted against the binary-searched boundary rather than a formula

Thanks to @mv-reyes, who caught during review that the first version of the same-token correction underflowed on any non-zero fee and returned `None` for exactly the positions it targeted. Every fixture here ran on a zeroed market at the time, so no test could see it; the sweep now carries fees in all eight shapes.

### Scope

Correlated but non-identical index and collateral tokens are deliberately left uncorrected. That error decays to zero as the position approaches liquidation and the UI recomputes continuously, so only genuinely identical tokens are handled here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,6 +65,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- sdk(sdk): Fixed `PositionStatus.liquidation_price`, which built its threshold from `min_collateral_factor` while a liquidation order is checked against `min_collateral_factor_for_liquidation`. The two differ on every mainnet market, so the reported price was wrong for every position regardless of collateral token; it read higher than reality for a long, warning earlier than the contract acts. The market-closed variant is covered too, since `position_params()` already resolves it.
 - programs(store): Reallocated the token map account with `zero_init` enabled.
 - programs(store): Forbade no-op swaps.
 - model: Treated a decrease that zeroes `size_in_tokens` as a full close.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - programs(store): An increase order now records its final output token at creation when the escrow is provided, which is what makes it eligible for a builder fee later.
 - sdk(solana-utils): Kept the two-argument `Bundle::send_all_with_opts` as a deprecated compatibility wrapper around the detailed API. It still returns the compressed success-signature list, and when multiple transactions fail it returns the **last** real send error (matching prior overwrite semantics; `SendAborted` placeholders are ignored).
 
+### Fixed
+
+- sdk(sdk): Fixed `PositionStatus.liquidation_price`, which built its threshold from `min_collateral_factor` while a liquidation order is checked against `min_collateral_factor_for_liquidation`. The two differ on every mainnet market, so the reported price was wrong for every position regardless of collateral token; it read higher than reality for a long, warning earlier than the contract acts. The market-closed variant is covered too, since `position_params()` already resolves it.
+- sdk(sdk): Fixed `PositionStatus.liquidation_price` when the collateral token is the market index token, where it held `collateral_value` and the pending funding fee at spot while solving for the very price both depend on. The correction is in the denominator only, since every other term is a plain USD amount; positions collateralised in a different token are unaffected. Correlated but non-identical tokens are deliberately left uncorrected, as that error decays to zero on approach to liquidation.
+
 ## [0.10.0] - 2026-08-12
 
 ### Breaking Changes
@@ -65,8 +70,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- sdk(sdk): Fixed `PositionStatus.liquidation_price`, which built its threshold from `min_collateral_factor` while a liquidation order is checked against `min_collateral_factor_for_liquidation`. The two differ on every mainnet market, so the reported price was wrong for every position regardless of collateral token; it read higher than reality for a long, warning earlier than the contract acts. The market-closed variant is covered too, since `position_params()` already resolves it.
-- sdk(sdk): Fixed `PositionStatus.liquidation_price` when the collateral token is the market index token, where it held `collateral_value` and the pending funding fee at spot while solving for the very price both depend on. The correction is in the denominator only, since every other term is a plain USD amount; positions collateralised in a different token are unaffected. Correlated but non-identical tokens are deliberately left uncorrected, as that error decays to zero on approach to liquidation.
 - programs(store): Reallocated the token map account with `zero_init` enabled.
 - programs(store): Forbade no-op swaps.
 - model: Treated a decrease that zeroes `size_in_tokens` as a full close.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,6 +66,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - sdk(sdk): Fixed `PositionStatus.liquidation_price`, which built its threshold from `min_collateral_factor` while a liquidation order is checked against `min_collateral_factor_for_liquidation`. The two differ on every mainnet market, so the reported price was wrong for every position regardless of collateral token; it read higher than reality for a long, warning earlier than the contract acts. The market-closed variant is covered too, since `position_params()` already resolves it.
+- sdk(sdk): Fixed `PositionStatus.liquidation_price` when the collateral token is the market index token, where it held `collateral_value` and the pending funding fee at spot while solving for the very price both depend on. The correction is in the denominator only, since every other term is a plain USD amount; positions collateralised in a different token are unaffected. Correlated but non-identical tokens are deliberately left uncorrected, as that error decays to zero on approach to liquidation.
 - programs(store): Reallocated the token map account with `zero_init` enabled.
 - programs(store): Forbade no-op swaps.
 - model: Treated a decrease that zeroes `size_in_tokens` as a full close.
--- a/crates/sdk/src/position/mod.rs
+++ b/crates/sdk/src/position/mod.rs
@@ -155,16 +155,17 @@ impl PositionCalculations for PositionModel {
         // the market leaves it unset, and `position_params()` already resolves the
         // market-closed variant, so reading it here covers both.
         let params = self.market().position_params()?;
-        let min_collateral_factor = params.min_collateral_factor_for_liquidation();
+        let min_collateral_factor_for_liquidation = params.min_collateral_factor_for_liquidation();
         let min_collateral_value = params.min_collateral_value();
-        let liquidation_collateral_usd = gmsol_model::utils::apply_factor::<
-            _,
-            { constants::MARKET_DECIMALS },
-        >(position_size_in_usd, min_collateral_factor)
-        .max(Some(*min_collateral_value))
-        .ok_or(gmsol_model::Error::Computation(
-            "calculating liquidation collateral usd",
-        ))?;
+        let liquidation_collateral_usd =
+            gmsol_model::utils::apply_factor::<_, { constants::MARKET_DECIMALS }>(
+                position_size_in_usd,
+                min_collateral_factor_for_liquidation,
+            )
+            .max(Some(*min_collateral_value))
+            .ok_or(gmsol_model::Error::Computation(
+                "calculating liquidation collateral usd",
+            ))?;
 
         // When the collateral token *is* the index token, two of the terms below are functions of
         // the very price being solved for, so they cannot be held at spot:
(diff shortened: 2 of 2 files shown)


mv-reyes:
Nice catch on the factor selection. I checked part 1 against the on-chain path (`check_liquidatable(.., true, true)` -> `min_collateral_factor_for_liquidation()`, the 0-fallback in `params/position.rs`, the market-closed resolution, and the max-with-`min_collateral_value` boundary) and it matches exactly.

One problem in the same-token correction, `crates/sdk/src/position/mod.rs:222-226`:

```rust
let fixed = if collateral_tracks_index {
    remaining_collateral_usd
        .checked_add(pending_funding_fee_value)?
        .checked_sub(collateral_value)?
```

`fixed` is the price-independent residual: `impact - borrowing_fee - close_order_fee`. Impact is clamped to <= 0 and both fees are >= 0, so `fixed` is never positive, and strictly negative whenever any fee is nonzero, which is every configured market (close order fee is always set). The `checked_sub` then returns `None` and `liquidation_price` comes out `None` for exactly the positions this commit targets (SOL/USD long collateralized in SOL). The test passes because the test market has all fees at zero; any real market config trips it.

u128-safe fix: reorder instead of going negative. `collateral_value - funding_fee - remaining = B + O - impact >= 0` unconditionally, so:

```rust
let fixed_complement = collateral_value
    .checked_sub(pending_funding_fee_value)?
    .checked_sub(remaining_collateral_usd)?;
// numerator_long = L + S + fixed_complement
```

Same idea for the short branch (`S + C - L - F - remaining`); if that one underflows, the position is already past the boundary and `None` is the honest answer there.

Disclosure on how this surfaced: I built a tool that does deterministic security review on Solana PR diffs (Anchor constraint classes, arithmetic, account lifecycle), and its checked-arithmetic ordering rule flagged this exact line as a low-confidence hint when I pointed it at your open PRs. I verified the algebra by hand before posting. The tool is public: github.com/mv-reyes/sol-pr-guard. It replays 24 real historical bugs from gmsol, drift, jito, klend and others offline (`npm run corpus`), including four from your own history, and zero dependencies, so a clone plus `npm test` is the whole setup.

The longer term goal is to grow it into something like a Copilot for Solana PR review, the thing that watches every PR so human reviewers can focus on logic instead of constraint bookkeeping. Constructive criticism and ideas on how to make it better are very welcome, especially from a team shipping a program of this size. If useful, point it at anything: `node dist/src/cli.js scan gmsol-labs/gmx-solana#439`.

Reyes

crypto1gmt:
Confirmed, and fixed in 890f31c4. Thank you, this was a real bug and the reasoning was exactly right.

Reproduction took one line: set `order_fee_factor_for_positive_impact` on an otherwise zeroed test market and `liquidation_price` comes back `None` for a long collateralized in the index token. Every test I had shipped ran on a fully zeroed market, so the residual was exactly 0 and the subtraction never went negative. The tests could not have caught this by construction, which is the part I should have noticed myself: I had already written down that the fee terms were untested and then treated the suite as if it covered them.

The fix carries your idea with one change. I ended up with `collateral - remaining - funding` rather than `collateral - funding - remaining`, because the first ordering is non-negative at both steps unconditionally:

```
collateral - remaining   = borrowing + funding + order_fee - impact   >= 0
...        - funding     = borrowing + order_fee - impact             >= 0
```

Your ordering needs `collateral >= funding` to hold at the intermediate step, which is true for any healthy position but is an assumption the other ordering does not need.

On the short branch I think the sign in your note is slipped. You wrote `S + C - L - F - remaining`, which expands to `S - L + surcharge`. Working it from the boundary I get `S - L - surcharge`:

```
remaining(P) = (ca - fa) * P + (impact - B - O)
short:  (ca - fa) * P + (impact - B - O) + S - T * P = L
        P * (T - ca + fa) = S - L - (B + O - impact)
```

I checked it rather than argue it. With `S - L - surcharge` the sweep matches the binary-searched `check_liquidatable` boundary to within 8 units on a ~1.8e12 price; flipping that one sign to `S - L + surcharge` moves it to 4444444441 units off. Same fixture, same boundary search, only the sign changed.

Also lifted your point into the tests: the sweep now runs all eight shapes (long/short, same/different collateral token, 0 and 50bps spread) with non-zero fees, so the residual is never 0 in any of them, and it asserts it reached all eight rather than skipping any.

On the tool: the interesting part for me is not that it found a line, it is that it found a line my tests structurally could not reach. A checked-arithmetic ordering rule does not need to understand the domain to notice that a subtraction is upstream of a value that cannot be negative, and that is a class humans and fixtures both miss. The honest limitation from this one data point is that it gave you a low-confidence hint and you did the algebra; the value came from the pair, not the rule alone. Whether we adopt anything is not my call, but I would be glad to keep giving you findings-level feedback on what it flags here.

mv-reyes:
You're right on both counts, and thank you for checking rather than arguing. The short-branch sign in my note was slipped; working it from the boundary gives S - L - surcharge, and your sweep numbers settle it. Your ordering is also strictly better: collateral - remaining funding is non-negative at both steps without needing collateral >= funding, so it holds for unhealthy positions too. I reasoned about the ordering when I should have run the sweep. Noted; that is exactly the gap the tool is meant to closeThe test-coverage point is the pattern I keep finding across repos: fixtures with everything zeroed make whole classes of arithmetic invisible, and the suite reads green anyway. Eight shapes with non-zero fees, plus asserting all eight actually ran, is the right fix.

On feedback, I will take you up on that. Two questions, since you have been deeper in this code than anyone reviewing from outside: if a tool like this earned a place in your review flow, what would it need to do that it does not do today? And what would it need to before the person whose call adoption is would trust it near the program side, not just the SDK? I would rather build the three things practitioners actually want than the thirty I can think of. @angelo-75 i would appreciate your input in this too please.

Meanwhile, the eight-shape sweep deserves to be generalized: I am putting a small PR that turns it into a differential harness, sweeping liquidation_price against check_liquidatable across randomized fee and impact configs, so the next ordering bug of this class fails loudly instead of shipping. Up in a day or two.

Thanks in advance for the input.

crypto1gmt:
One thing from today, since it cuts against the rule rather than for it.

I tried to improve the ordering you endorsed: `borrowing + order_fee + |impact|` instead of `collateral - remaining - funding`. Same value, no subtraction, sweep green. Then I reverted it: the subtractive form tracks the chain producing `remaining`, the additive one does not. Add any price-independent `X` and the correct surcharge becomes `B + O + X - I`, which the subtraction yields on its own and the addition drops silently, suite still green. So the rule that got the original right would flag the correct code and push the worse fix. A false positive with teeth.

Three things I would want, in order:

1. **Sign provenance.** Your hint was low confidence because the tell sat ~140 lines upstream (`if x.is_negative() { .. } else { x = Zero::zero() }`). Propagate "known non-positive" forward and it fires as a finding, not a hint.
2. **Fixture reachability.** Nothing caught the original because every fixture was `Market::zeroed()`. "Config pins `F` at 0, branch needs `F != 0`, so this test cannot reach it" is mechanical and catches the class before the bug.
3. **A trustworthy negative.** Ship every sweep with a positive control in the same pass. I just shipped a check whose comparison was lexical, not numeric: it printed the two correct line numbers and would have flagged almost anything.

Adoption on the program side is not my call. Factually: defects there are lifecycle, not arithmetic, a write or an event landing before a fallible step inside an error-swallowing region. And doc comments compile into the IDL, so a stale one is a shipped-artifact defect.

On the harness: assert which shapes it actually reached, not just that assertions passed. With randomized configs a region can go ungenerated while the run stays green.

CHANGELOG.md
```diff
@@ -39,6 +39,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - programs(store): An increase order now records its final output token at creation when the escrow is provided, which is what makes it eligible for a builder fee later.
 - sdk(solana-utils): Kept the two-argument `Bundle::send_all_with_opts` as a deprecated compatibility wrapper around the detailed API. It still returns the compressed success-signature list, and when multiple transactions fail it returns the **last** real send error (matching prior overwrite semantics; `SendAborted` placeholders are ignored).
 
+### Fixed
+
+- sdk(sdk): Fixed `PositionStatus.liquidation_price`, which built its threshold from `min_collateral_factor` while a liquidation order is checked against `min_collateral_factor_for_liquidation`. The two differ on every mainnet market, so the reported price was wrong for every position regardless of collateral token; it read higher than reality for a long, warning earlier than the contract acts. The market-closed variant is covered too, since `position_params()` already resolves it.
```
angelo-75: suggestion: The second sentence, "The two differ on every mainnet market ...", should be avoided, since this repository does not make assumptions about the deployment state. Consider removing the sentence entirely, or rewording it so it does not depend on the deployment state, for example: "markets typically set the two to different values..."

crates/sdk/src/position/mod.rs
```diff
@@ -198,3 +285,557 @@ impl PositionCalculations for PositionModel {
...
+            .status(&prices())
+            .expect("status")
+            .liquidation_price
+            .expect("liquidation price");
+        assert_eq!(reported, expected(mcf));
+    }
+
+    /// A market whose index token IS its long token, which is the shape the correction exists
+    /// for. The assertion is not against a formula of my own: it binary-searches the real
+    /// `check_liquidatable(.., true, true)` and requires the reported price to be that boundary.
```
angelo-75, line 411: nit: Who does "my own" refer to? Please use objective wording in all comments.

angelo-75 (dismissed):

CHANGELOG.md
crypto1gmt: Reworded in c0c6a2d5: "Whenever a market sets the two to different values, ..." so the entry no longer depends on the deployment state.

crates/sdk/src/position/mod.rs
crypto1gmt: Fixed in 6abb2537. Also checked the rest of the diff for first-person wording, this was the only instance.

angelo-75 (approved):

mv-reyes:
Glad to see this one land. All three are in.

- Sign provenance: the ordering rule now propagates known-non-positive forward through enforced top-level facts (require!/assert!/require_gte!, reject-with-abort), and a hint on the exact chain value (or an alias bound to it) promotes to a finding. Conditional or closure facts stay hints; reassignment or a non-readonly call voids the fact.
- Fixture reachability: new rule cross-checks config pins against branch conditions across files and flags the "config pins F at 0, branch needs F != 0" shape. Runs as T2.
- Trustworthy negatives: every sweep now ships positive controls (7 known-buggy fixtures through the real engine); a control miss fails the run instead of printing an all-clear.

On the program side: agreed, the defects that matter there are lifecycle, not arithmetic. The early-return-skips-cleanup class (a write or event landing before a fallible step inside an error-swallowing region) is the drift pattern I already have corpus entries for, so that is the next rule family. Noted on doc comments compiling into the IDL; a stale one is a shipped artifact, so the IDL surface is in scope.

The differential harness is done: 172 stratified cases of liquidation_price vs check_liquidatable(.., true, true), binary-search oracle with a per-case positive control, and a reachability ledger that fails the run on under-coverage (it asserts which shapes were actually reached, per your spec). Seed 20260915: 160 priced cases, worst error 28 bps on a 50bps-spread shape; zero-spread cases land exactly on the boundary. PR against the sdk crate coming next.

Suggestions and constructive criticism very welcome, on the tool or the harness. The goal is something worth having watch every PR, and you know where the bodies are buried in this codebase better than anyone.

Reyes

mv-reyes:
Harness PR: https://github.com/gmsol-labs/gmx-solana/pull/443

PR #440, opened 2026-09-14
Add v1 transaction decoder and tests for Solana transactions

hhhhhpp:


examples/decode_transaction.rs
```diff
@@ -138,7 +138,7 @@ async fn main() -> gmsol_sdk::Result<()> {
                 RpcTransactionConfig {
                     encoding: Some(UiTransactionEncoding::Base58),
                     commitment: None,
-                    max_supported_transaction_version: Some(0),
+                    max_supported_transaction_version: Some(1),
```
crypto1gmt, line 141: The RPC refuses this pair: `Base58` two lines up plus `Some(1)` gives `-32602 base58 encoding is not supported with maxSupportedTransactionVersion >= 1`. Checked on devnet, size-independent, so every call from this example fails. `transaction_history.rs` already forces `Base64`, maybe here too.

crates/decode/src/decoder/decoder_impl/fixtures/README.md
```diff
@@ -0,0 +1,7 @@
+# V1 reference fixtures
+
+`v1-reference.json` holds 20 synthetic cases, one per line, serialized and sanitized by the official Anza codec in an isolated project. Each case includes the original expected fields. Signatures and test RPC meta are synthetic; real provider replay is still required before deployment.
```
crypto1gmt, line 3: The wire parse looks well covered, since the generator named below is the same codec the RPC serializes with. The thinner part is the envelope: that `getTransaction` returns the `0x81` bytes as base64, and that `meta.loaded_addresses` really is absent for v1, which the decoder hard-errors on. Round-tripping one transaction on devnet would close both.

crypto1gmt (approved):

angelo-75 (approved):

PR #441, opened 2026-09-14
feat(sdk): add atomic-group builders for the remaining builder fee instructions

crypto1gmt:
Three of the four v0.11 builder fee instructions existed only in the client ops path and had no atomic-group builder: `settle_builder_fee` and `claim_builder_fees` on `BuilderFeeOps`, and `set_builder_fee_factor` on `UserOps`. `set_builder_fee` already had one, so it is the shape the three follow: `impl IntoAtomicGroup` with an explicit `Hint`.

This blocks the NPM surface rather than sitting beside it. The JS layer is built on the builder layer, so a `wasm_bindgen` wrapper has nothing to call until these exist.

### Builders

- `ClaimBuilderFees` in a new `builders/builder_fee.rs`, not under `order/`: it takes no order, only the store, the mint and a destination. `Hint = ()`, since every account is derived from those plus the payer.
- `SetBuilderFeeFactor` in `builders/user.rs`, next to `PrepareUser`. Also `Hint = ()`.
- `SettleBuilderFee` in `builders/order/settle_builder_fee.rs`, with a hint resolvable from the order account through `FromRpcClientWith`.

### Breaking change

`BuilderFeeOps::settle_builder_fee` now takes the builder layer's `SettleBuilderFeeHint`, and the identically named type beside the trait is removed. Field names and meanings are unchanged; they hold `StringPubkey` rather than `Pubkey`. Callers passing `None` are unaffected, which is every call site in this repo. This mirrors what `set_builder_fee` already did, so the two neighbouring methods no longer disagree about where their hint lives.

### Settlement stays a no-op when nothing is owed

The zero-amount path attaches none of the four optional accounts, so they need not exist on-chain, and the three missing-field errors each name themselves rather than sharing one message. Both behaviours are ported verbatim from the ops implementation and pinned by tests.

### Tests

Five, each written to fail for one specific substitution: the claim vault derived from the owner key instead of the User Account, the factor written to the owner instead of the User Account PDA, the settlement claim vault derived from the payer instead of the builder, the no-op path attaching accounts it must not, and the three error messages collapsing into one.

### Out of scope

Composing `settle_builder_fee` into the close order flow is deliberately not here. It is a separate design point still being settled, and it changes the shape of this builder rather than adding to it.


crates/sdk/src/builders/order/settle_builder_fee.rs
```diff
@@ -0,0 +1,326 @@
...
+                    claim_vault,
+                    token_program: anchor_spl::token::ID,
+                    event_authority: self.program.find_event_authority_address(),
+                    program: self.program.id.0,
+                },
+                &ID,
+                &self.program.id.0,
+            ),
+            data: args::SettleBuilderFee {}.data(),
+        };
```
angelo-75: issue: You should build the instruction with `self.program.anchor_instruction(...).anchor_accounts(...).build()`, which comes with built-in handling for optional accounts.

crates/sdk/src/builders/order/settle_builder_fee.rs
```diff
@@ -0,0 +1,326 @@
...
+    /// caller-supplied hint impossible to debug. Each has to name itself.
+    #[test]
+    fn each_missing_hint_field_names_itself() {
+        let payer = Pubkey::new_unique();
+        let order = Pubkey::new_unique();
+        let some = || Some(StringPubkey::from(Pubkey::new_unique()));
+
+        let cases = [
+            (
+                "no final output token mint",
```
angelo-75, line 231: nit: Consider using constants for the error messages to avoid repeating the hard-coded strings.

crates/sdk/src/builders/builder_fee.rs
```diff
@@ -0,0 +1,152 @@
...
+                    claim_vault,
+                    destination: self.destination.0,
+                    user_token_controller: self
+                        .program
+                        .find_user_token_controller_address(&user_account, &token_mint),
+                    token_program: token_program_id,
+                    event_authority: self.program.find_event_authority_address(),
+                    program: self.program.id.0,
+                },
+                true,
```
angelo-75: nit: When there are no optional accounts, there is no need to pass `true`, which saves some computation.

crates/sdk/src/builders/user.rs
```diff
@@ -45,3 +45,95 @@ impl IntoAtomicGroup for PrepareUser {
...
+            })
+            .anchor_accounts(
+                accounts::SetBuilderFeeFactor {
+                    owner,
+                    store: self.program.store.0,
+                    user: self.program.find_user_address(&owner),
+                    event_authority: self.program.find_event_authority_address(),
+                    program: self.program.id.0,
+                },
+                true,
```
angelo-75: nit: When there are no optional accounts, there is no need to pass `true`, which saves some computation.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Breaking Changes
 
 - programs(store): Creating an increase order now requires its final output token to be the position's collateral token, and executing one whose final output token was recorded at creation revalidates the same thing. Creating an order with a different final output token used to succeed and silently ignore the value; it now reverts with `TokenMintMismatched`. Existing orders with an uninitialized final output token are unaffected and keep executing.
+- sdk(sdk): `BuilderFeeOps::settle_builder_fee` now takes the builder layer's `SettleBuilderFeeHint`, and the identically named type that used to sit beside the trait has been removed. Field names and meanings are unchanged; they now hold `StringPubkey` instead of `Pubkey`. Callers passing `None` are unaffected. This mirrors what `set_builder_fee` already did, so the two no longer disagree about where their hint lives.
```
angelo-75: issue: Changes to an unreleased feature do not count as breaking changes.

angelo-75 (changes requested):

crates/sdk/src/builders/order/settle_builder_fee.rs
crypto1gmt: Done, switched to `anchor_instruction().anchor_accounts(..., true).build()`.

crates/sdk/src/builders/order/settle_builder_fee.rs
crypto1gmt: Done, extracted to `ERR_NO_MINT`, `ERR_NO_ESCROW`, `ERR_NO_BUILDER`.

crates/sdk/src/builders/builder_fee.rs
crypto1gmt: Done, changed to `false`.

crates/sdk/src/builders/user.rs
crypto1gmt: Done, changed to `false`.

CHANGELOG.md
crypto1gmt: Done, moved to `### Changed`.

angelo-75: It is not a change either, since it did not exist in v0.10.

angelo-75 (approved):

PR #442, opened 2026-09-15

crypto1gmt:
## Summary

Depends on #441.

## Test plan

- [x] `cargo check --features js,serde` passes (verified locally)
- [ ] CI passes
- [ ] Types generated by `wasm-pack build` include the new structs and functions


crates/sdk/src/js/instructions/close_order.rs
```diff
@@ -30,6 +30,12 @@ pub struct CloseOrderArgs {
     program: Option<StoreProgram>,
     #[serde(default)]
     transaction_group: TransactionGroupOptions,
+    /// Per-order `settle_builder_fee` hints, keyed by order address.
+    ///
+    /// When present, a `settle_builder_fee` instruction is emitted in a stage
+    /// before the close instructions so the escrow still exists when it runs.
+    #[serde(default)]
+    settle_builder_fee: HashMap<StringPubkey, SettleBuilderFeeHint>,
```
angelo-75: suggestion: This is a breaking change, because it is not wrapped in an Option (wasm-bindgen handles defaults slightly differently: if the type is not an Option, it remains required). But even if we add the Option so that it is no longer a breaking change, it still provides a mitigation for the breaking change introduced by the builder fee, so it is worth mentioning in the changelog. Specifically: callers of close order should check for and provide the necessary settle builder fee hints, to ensure that close order can always go through.

crates/sdk/src/js/instructions/create_order.rs
```diff
@@ -54,6 +66,15 @@ pub struct CreateOrderOptions {
...
     force_create_positions: Option<bool>,
+    /// Per-order `set_builder_fee` options, keyed by market token.
+    ///
+    /// When an entry is present for an order's market token, a `set_builder_fee`
+    /// instruction is appended in the same transaction group (after all create-order
+    /// instructions). The order address is derived internally, so the `nonce` field
+    /// in [`CreateOrderParams`] is optional here. Set it if you need the order
+    /// address before sending; the returned `TransactionGroup` does not expose it.
+    #[serde(default)]
+    set_builder_fee: HashMap<StringPubkey, SetBuilderFeeOptions>,
```
angelo-75: issue: This is a breaking change. Also, it does not really make sense to have different builder settings per market, so we should consider simplifying it to a single optional builder hint. That way it is no longer a breaking change, and integrators do not have to "expand" these per-order builder hints themselves.

angelo-75 (changes requested):

angelo-75:
By the way, please consider running or adding end-to-end tests in crates/sdk/tests/web/demo.ts. That helps catch these breaking changes.

crates/sdk/src/js/instructions/close_order.rs
crypto1gmt: Done. Wrapped in `Option<HashMap<...>>`, usage updated to `unwrap_or_default()`. Added a changelog note advising callers to populate the field when closing orders that may carry a non-zero fee.

crates/sdk/src/js/instructions/create_order.rs
crypto1gmt: Done. Simplified to `Option<SetBuilderFeeOptions>`: same settings apply to every order in the call. Updated the changelog accordingly.

crypto1gmt:
Added e2e smoke tests for the builder-fee fields in demo.ts, commit 4cfcdfdc.

crypto1gmt:
Superseded by #445: GitHub would not let this PR be reopened once its base branch (`feat/builder-fee-atomic-group-builders`) was auto-deleted on #441 merging. Same commits, rebased onto `main`.

PR #443, opened 2026-09-16
test(sdk): differential sweep of liquidation_price vs check_liquidatable

mv-reyes:
Follow-up to the #439 discussion: this adds the differential harness we talked about.

What it does:
- 172 stratified cases across long/short x same/different collateral x 0/50bps spread x impact shape (5 seeded instances each, plus 12 None shapes).
- The oracle is a binary search on the real check_liquidatable(.., true, true) with a per-case positive control, so a case only counts when the oracle actually crosses the boundary.
- A reachability ledger per shape: the run fails loudly on under-coverage instead of silently passing (impact sign is measured, not assumed).

Current numbers, seed 20260915: 172 cases, 160 priced, worst observed error 28bps on a 50bps-spread case; zero-spread cases land exactly on the boundary.

Dependency note: one dev-dependency, sol-rpc-mini (git, rev-pinned), reused from my own tooling for its det module: a spec-seeded RNG and deterministic fixture pubkeys. The harness never touches the network. It is a dev-dependency only, so none of it ships in the SDK. The crate does pull transitive deps via its RPC half (ureq/rustls); if that is unwanted here I can inline the ~120 lines of det instead.

Run: cargo test -p gmsol-sdk --test liquidation_differential

Test-only; nothing in the SDK build changes. Merge it, adapt it, or keep it as a reference, whatever fits your process.


crypto1gmt:
Reviewed against the three properties you named, and actually ran it rather than just reading it: cargo test -p gmsol-sdk --test liquidation_differential, all green locally. sweep: 172 cases, 160 priced, worst error 28bps (case-103), matches your PR description exactly.

Worth flagging first: your CI has never actually run. Both workflows are stuck at GitHub's action_required gate for a first-time fork contributor, not failed, never started. Someone with write access needs to approve the pending run before this has any signal beyond local.

Sign provenance: confirmed, and I tried to break it. Removed ImpactShape::Negative from the case generator expecting to starve the impact/negative bucket. It didn't, because Neutral's cross-over path also produces a negative measured impact, exactly what "provenance from measurement, not config" predicts.

Fixture reachability: confirmed by deliberately breaking it. Restricted is_long to [true] only, the test failed exactly as designed: side/short: 8 reached, 30 required.

Positive controls: confirmed, and stronger than the precedent. #439's own sweep_reported_vs_real_boundary only checks the search endpoints; yours checks check_liquidatable right at the found boundary, which is a real improvement, not just a restatement.

Two minor things:

nit: Fixture::liquidation_collateral_usd is dead, cargo warns on it.

suggestion: sol-rpc-mini pins with git = "..." but no rev =, so the pin only lives in Cargo.lock. Looked at what inlining would actually take: det.rs is 124 lines like you said, but fixture_bytes also pulls in a 117-line table submodule (serde_json, env var override, file loading) for cross-release fixture stability, which a same-repo test doesn't need. The honest inline is smaller than your estimate once that's dropped, more reason to take you up on it, not less.

mv-reyes:
Appreciate the adversarial pass, especially the two break attempts. That is exactly the review the harness was built to survive.

The nit is fixed: `liquidation_collateral_usd` was a leftover from an earlier collateral-sizing shape. Removed. The manifest also pins the rev directly now instead of relying on the lockfile alone. Both in f03c8f5.

On inlining, I went back and forth after your message and land on keeping the dependency. The 117-line table submodule is the reason: it is the fixture-format loader, not incidental weight, and an inline would fork it the first time either side changes the format. What I did take from your point: 0.3.1 just landed upstream, moving the seed rows to a compile-time table (a small `det-macros` proc-macro crate) with built-in fallback rows, so the labels survive a missing or unreadable asset. That makes the format more versioned, not less, and the harness should exercise the loader and fallback the crate ships rather than a copy of it. The bump in this push moves the harness to it.

CI: thanks for flagging, I read it as failed rather than never-started. Pending the `action_required` approval whenever someone with write access gets to it.

PR #445, opened 2026-09-18

crypto1gmt:
## Summary

Rebased onto `main` after #441 merged. Supersedes #442, which GitHub would not let us reopen once its base branch (`feat/builder-fee-atomic-group-builders`) was auto-deleted on merge. Same commits, replayed with `git rebase --onto`, no conflicts.

## Test plan

- [x] `cargo check --features js,serde` passes (verified locally)
- [x] Rebase re-verified locally: `cargo check -p gmsol-sdk` (native) and `--features js --target wasm32-unknown-unknown` both pass clean against the merged #441
- [ ] CI passes
- [ ] Types generated by `wasm-pack build` include the new structs and functions


mv-reyes:
Good simplification moving `set_builder_fee` off the per-market map, but I think it reintroduces a bug.

`create_orders_builder` now applies one shared `set_builder_fee.final_output_token` to every order in the call (the `SetBuilderFeeHint` built inside the map closure in create_order.rs). The rest of the API still supports multi-market batches, with `hints` keyed by market token, but the fee hint is a single token for all of them. In a batch spanning two markets, every order but one gets a hint naming the wrong output token. If the on-chain `set_builder_fee` validates the recorded token against the order's actual output, those instructions fail and take the whole transaction group down with them. If it does not validate, the fee is booked against the wrong mint and `claim_builder_fees` settles against the wrong accounting later.

The changelog warns that builder and factor are shared across the call but never mentions the token, so callers will not see this coming. The PR body still describes the per-market map, which makes me think the map was the intent.

Fixes, cheapest first:

- restore the map keyed by market token, or
- derive the output token per order from the swap path and market hint, and drop it from the options, or
- if single-market batches are the intent, reject multi-market calls when `set_builder_fee` is set so it fails loudly in the SDK instead of on-chain.

Minor while you are in there: `HashMap` iteration in the settle and close stages makes the emitted transaction bytes nondeterministic for identical inputs. Harmless on-chain, but it will flake any byte-level snapshot tests on the JS side.

Unrelated: the things from the #439 thread are in #443 now. Sign provenance propagating through enforced facts, fixture reachability as a T2 rule, and every sweep shipping positive controls, per your spec. If one of you could give it a pass when you have a minute, I want feedback on scope and on anything you would add or remove. I am shaping this toward a copilot for Solana review rather than a one-off harness, and direction from people who maintain a program this size is what I need most.

crypto1gmt:
Superseded by #447: this PR is from the fork, which never gets repo secrets on a plain pull_request run, exactly the SCDEV/CON write-access issue from early September resurfacing. Opened #447 from a branch pushed directly to this repo instead, same commits plus the two fixes below, and its CI runs clean including test-programs.

PR #447, opened 2026-09-18

crypto1gmt:
## Summary

Two fixes on top of the original #442/#445 content, from review:
- `create_orders_builder` wired `set_builder_fee` onto increase orders without ever opting them into the final-output-token escrow, so the resulting `SetBuilderFee` was silently ineffective. Fixed by deriving the escrow's target token from the builder fee options when the caller doesn't already specify one, and letting `CreateOrder`'s increase-order branch actually use it instead of hardcoding `None`.
- `tests/web/demo.ts` passed `builder_fee_amount` as a `bigint` literal where `SettleBuilderFeeHint` expects `number`, so the demo never actually built. Fixed, verified with `npm run build`.

Supersedes #442 and #445, both closed: their head branch was on the fork, which gets no repo secrets on a plain `pull_request` run, so `test-programs` could never pass there regardless of the diff. This branch is pushed directly to this repo instead.

## Test plan

- [x] `cargo check --features js,serde` passes
- [x] `cargo check -p gmsol-sdk --target wasm32-unknown-unknown --features js` passes
- [x] `npm run build` (webpack + wasm-pack) passes clean, including `tests/web/demo.ts`
- [x] CI passes (test-programs green from a same-repo branch)


mv-reyes:
The shared-token point from #445 still applies here, and the escrow fix extends it. `create_orders_builder` applies the single `set_builder_fee.final_output_token` to every order, and the new `receive_token` fallback now feeds that same token into the final-output escrow opt-in for every increase order in the call. Per your own changelog, an increase order records its final output token at creation when the escrow is provided — so in a multi-market batch, every order but one records the wrong output token on-chain at creation, not just in the fee hint.

One more asymmetry in the same area: the `if let Some(sbf_opts)` block fires for every order kind, but the escrow opt-in only happens when `kind.is_increase()`. A mixed batch, or a swap-only call with `set_builder_fee` set, emits fee instructions for orders that never opted into an escrow. If the program expects the final-output record to exist, the group reverts; if it does not, the fee attaches to orders where it means nothing. Does the program side reject `set_builder_fee` on swaps and decreases, or should the SDK gate it?

And the standing ask, once, then I will stop repeating it: the three things you specced in #439 are in #443 — sign provenance, fixture reachability, positive controls on every sweep. If it is worth your time, give it a pass and tell me what to add or remove. If it is not, no hard feelings, and I will stop bringing it up.

crypto1gmt:
Good catches on both, thanks.

**Multi-market batch**: right, `set_builder_fee.final_output_token` was getting fed into `receive_token` for every order in the call, so every order but the one actually matching that token would record the wrong output token on-chain at creation, not just in the settle hint. Now rejecting early when more than one order is passed alongside `set_builder_fee`, since the API is already documented as single-builder/single-factor per call; this makes it single-order too instead of silently doing the wrong thing for a batch.

**Swap gating**: checked the program side, `set_builder_fee` allows increase and decrease orders (`OrderKind::is_user_initiated_position`) but rejects swaps with `BuilderFeeOrderKindNotAllowed`. Decrease orders already get an unconditional escrow in the existing code, so they were never actually broken, only swaps were, since the SDK would happily build a transaction group that's guaranteed to revert on execution. Now rejecting swap + `set_builder_fee` at build time in the SDK too, same reasoning.

Both pushed, verified with `cargo check` (native + wasm) and `npm run build`.

On #443: haven't gone through it against the #439 spec yet, will get to it.

CHANGELOG.md
```diff
@@ -39,6 +39,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - programs(store): An increase order now records its final output token at creation when the escrow is provided, which is what makes it eligible for a builder fee later.
+- sdk(js): `CreateOrderOptions::set_builder_fee` simplified from a per-market `HashMap<marketToken, SetBuilderFeeOptions>` to a single `Option<SetBuilderFeeOptions>`. The same builder and factor apply to every order in the call; callers needing different settings per order should use separate `create_orders_builder` calls.
```
orion-047: issue: the changelog is relative to the last released version, so please remove this entry (and please have the AI remember this).

CHANGELOG.md
```diff
@@ -39,6 +39,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- sdk(js): `CloseOrderArgs::settle_builder_fee` is now `Option<HashMap<orderAddress, SettleBuilderFeeHint>>`. Callers that do not provide it get the same behaviour as before; callers closing orders that may carry a non-zero builder fee should populate it so the fee is settled before the escrow is closed.
```
orion-047: issue: the changelog is relative to the last released version, so please remove this entry (and please have the AI remember this).

crates/sdk/tests/web/demo.ts
```diff
@@ -549,3 +549,166 @@ for (const batch of glvWithdrawalGroup.serialize()) {
     console.log(toBase64(txn));
   }
 }
+
+// Builder fee: close_orders with settle_builder_fee populated.
+const builderPubkey = "22222222222222222222222222222222";
+const finalOutputToken = wsol;
+const escrowPubkey = "33333333333333333333333333333333";
+const closeOrdersWithFee = close_orders({
```
orion-047: issue: `npm run start` starts failing at this line, presumably because the two addresses above are malformed. Consider replacing them with valid base58 addresses.

crates/sdk/src/js/instructions/close_order.rs
```diff
@@ -73,6 +79,20 @@ pub fn close_orders(args: CloseOrderArgs) -> crate::Result<TransactionGroup> {
...
+        .into_iter()
+        .map(|(order, hint)| {
+            Ok(SettleBuilderFee::builder()
+                .payer(payer)
+                .order(order)
+                .program(program.clone())
+                .build()
+                .into_atomic_group(&hint)?)
+        })
+        .collect::<crate::Result<ParallelGroup>>()?;
```
orion-047: issue: Settle should not be its own parallel group. Downstream consumers don't necessarily honor the ordering contract of transaction groups, so the settle group could end up being sent in parallel with the close group. Consider merging each order's settle and close into a single atomic group. Also, a small technical decision here: if settle hints are provided, their shape should match orders exactly, i.e. the keys of `orders` and `settle_builder_fee` must be the same set.

crates/sdk/src/js/instructions/create_order.rs
```diff
@@ -63,6 +83,23 @@ pub fn create_orders_builder(
...
+                 (BuilderFeeOrderKindNotAllowed), only increase and decrease orders may carry one",
+            ));
+        }
+        if orders.len() > 1 {
+            return Err(crate::Error::custom(
+                "set_builder_fee applies one final_output_token to every order in the call; a \
+                 multi-order batch would record it as every order's own output token at creation, \
+                 which is only correct for orders that actually share it. Use separate \
+                 create_orders_builder calls per order instead",
+            ));
```
orion-047: issue: `receive_token` is semantically the same as `final_output_token`, so there's no need to provide it separately in the hint. In fact, this instruction builder assumes that all orders share the same pay token and the same receive token respectively (i.e. the final output token is the same across them). The only case to watch is increase orders: previously receive_token was hardcoded to None, but you've changed it to be passed through, so in the `set_builder_fee` branch, `receive_token` should be set to the collateral token.

crates/sdk/src/js/instructions/create_order.rs
```diff
@@ -215,6 +287,7 @@ impl CreateOrdersBuilder {
                 .add(prepare)?
                 .add(self.positions.into_values().collect::<ParallelGroup>())?
                 .add(self.create.into_iter().collect::<ParallelGroup>())?
+                .add(self.set_builder_fees.into_iter().collect::<ParallelGroup>())?
```
orion-047: issue: `set_builder_fee` must immediately follow create and be in the same atomic group; splitting them is not allowed. See SCSOL-13 for the rationale.

orion-047 (changes requested):

CHANGELOG.md
crypto1gmt: Removed in 06161118. Same for the `settle_builder_fee` entry below, neither API existed in v0.10 so there is nothing to have changed.

CHANGELOG.md
crypto1gmt: Removed in 06161118, along with the one above.

crates/sdk/tests/web/demo.ts
crypto1gmt: Fixed in 474f0c7b. They decoded to 23 bytes rather than 32, no leading 1s to supply the zero bytes. Swapped for 32-byte fills of 0x22 and 0x33, and checked every other base58 literal in the file while I was in there.

crates/sdk/src/js/instructions/close_order.rs
crypto1gmt: Done in d528934f. Each order's settle and close now go in one atomic group, settle first, so the ordering is instruction order inside a single transaction rather than a contract the consumer can ignore. Also added the key check you mentioned: `settle_builder_fee` and `orders` have to be the same set, and the error names what is missing and what is extra.

crates/sdk/src/js/instructions/create_order.rs
crypto1gmt: Done in ba294ee3. `set_builder_fee` now merges into the order's own atomic group right after create. The separate stage is gone entirely, so this one deletes a code path rather than adding one.

crates/sdk/src/js/instructions/create_order.rs
crypto1gmt: Two halves here and I want your call on the second.

The increase branch is clear: `receive_token` should be the collateral token, not the fee's `final_output_token`. I can do that now.

Dropping `final_output_token` from `SetBuilderFeeOptions` is the one I paused on. It is wasm-exported, so the TS surface and `demo.ts` move with it, and the multi-order guard above is argued from that exact field ("applies one `final_output_token` to every order in the call"). Derive the token from `receive_token` instead, which is already shared across the batch, and that guard loses its premise, so it would need reworking or removing. It went in for a bug mv-reyes found, so I did not want to undercut it on a guess. Field gone and guard reworked, or just the increase branch?

mv-reyes:
Small follow-up on the settle+close grouping in d528934f. Non-blocking, but flagging while the code is still open.

The JS entry points take `SettleBuilderFeeHint` caller-supplied, and the builder only validates field presence (the named-error tests pin this), not freshness. Reasonable, since it never touches the chain. But every field is load-bearing on-chain: a stale `escrow` trips `TokenAccountMismatched`, a stale `builder` trips `InvalidUserAccount`, and a `builder_fee_amount` read while the order was still pending (zero) that is non-zero by submission builds the no-op settle, which reverts with `TokenMintNotProvided`. Since close requires the fee settled and the merge reverts the two together, the realistic sequence of page loads while pending, keeper executes, user closes with the stale hint turns into the whole close failing until hints are refreshed. State is never at risk; it's a liveness trap for the caller.

The Rust side already has the answer in `FromRpcClientWith<SettleBuilderFee>`, which derives the hint from the live order account. The JS surface has no equivalent, so the fix on your side is either a small exported helper that refreshes hints from the order account right before `build_with_options`, or a doc note on `settle_builder_fee` / `CloseOrderArgs.settle_builder_fee` saying hints must come from a fresh order fetch.

One-line nit while in there: the doc on `SetBuilderFeeOptions` says "keyed by the order's market token", but the field is a single `Option`, not a map. Leftover from the earlier shape.

(Unrelated: #443 is still open whenever review bandwidth frees up. The nit is fixed, `liquidation_collateral_usd` was a leftover from an earlier collateral-sizing shape and is removed.)

PR #448, opened 2026-09-21
Add CLI commands for the builder fee instructions

crypto1gmt:
The CLI carries no commands for `set_builder_fee`, `set_builder_fee_factor` or `claim_builder_fees`, so exercising any of them means writing code against the SDK directly. `BuilderFeeOps` (for the first and third) and `UserOps::set_builder_fee_factor` already exist and are unchanged; this PR only wires the CLI to them.

### Commands

All three added to `crates/cli/src/commands/user.rs`, alongside the other self-service, own-account operations (referral codes):

- `user set-builder-fee <ORDER> --builder <BUILDER> --expected-factor <FACTOR>`
- `user set-builder-fee-factor <FACTOR>`
- `user claim-builder-fees <TOKEN_MINT> [--destination <DESTINATION>]`

`--expected-factor`/`<FACTOR>` take the same `Value` decimal notation as every other factor-typed CLI argument (e.g. `configuration insert-factor`), converted with `.to_u128()` before the SDK call. The help text spells that out, because the value has to match what the builder advertises on-chain and `inspect account` prints it as a raw integer. `set_builder_fee`'s hint is always resolved from the order account (`None` passed through), matching every other hinted command in this file.

`claim-builder-fees` sits under `user` rather than beside an existing `Claim*` command: `treasury claim-fees` batches protocol market fees into the receiver vault, `lp claim-gt` is scoped to an LP staking position, and neither shares this command's shape (an arbitrary caller draining their own builder claim vault for one mint). Mechanically it derives everything from the caller's own User Account exactly like the referral-code commands beside it, so it stays here rather than starting a new top-level group.

`--destination` defaults to the payer's associated token account for `TOKEN_MINT`, merging an idempotent `create_associated_token_account_idempotent` ahead of the claim. Every other destination-shaped argument in the CLI is already optional, and the program never creates the destination itself, so without the default a builder who has never held the fee mint is blocked on an account `gmsol` has no command to create. An explicitly passed `--destination` is used as-is and never created, so this never creates an account for a third party.

Beyond that derivation, every command is a direct pass-through to the existing ops, so a failing instruction surfaces through the SDK's existing `Error::Anchor { error_name, .. }`, decoded from the simulation log, the same as every other command in the file, not a raw error code.

### Out of scope

`settle_builder_fee` has no CLI command either. The order stays stuck with no command able to close it, which is a distinct bug already tracked as [CON-62](https://linear.app/gmtradepro/issue/CON-62/cli-cannot-settle-or-close-a-builder-fee-order-after-execution-no) rather than a coverage gap this PR should also pick up.


crates/cli/src/commands/user.rs
```diff
@@ -24,6 +27,40 @@ enum Command {
...
+        /// The order to attach the builder fee to.
+        order: Pubkey,
+        /// The builder receiving the fee.
+        #[arg(long)]
+        builder: Pubkey,
+        /// Expected fee factor, in decimal notation (`0.001` is 0.1%), not the raw
+        /// on-chain integer. Must equal what `builder`'s User Account currently
+        /// advertises; the call is rejected on any mismatch.
+        #[arg(long)]
+        expected_factor: Value,
```
orion-047: nit: We could make this parameter optional to improve the user experience.

orion-047: This may mean we need to make `expected_factor` part of the hint. So I'd suggest not handling it in this PR (otherwise the PR's scope would become too large).

crates/cli/src/commands/user.rs
```diff
@@ -57,6 +94,36 @@ impl super::Command for User {
...
+            }
+            Command::ClaimBuilderFees {
+                token_mint,
+                destination,
+            } => match destination {
+                Some(destination) => client.claim_builder_fees(store, token_mint, destination)?,
+                None => {
+                    // The claim vault is an associated token account of the legacy
+                    // token program, so the mint is one too and the default
+                    // destination derives unambiguously.
```
orion-047: issue: The SPL Token program is not legacy.

orion-047 (changes requested):

crates/cli/src/commands/user.rs
crypto1gmt: Not handling it here, per your note. Agreed it pulls `expected_factor` into the hint, which is more than this PR should carry.

crates/cli/src/commands/user.rs
crypto1gmt: Fixed in 58c6c1d9. Checked the rest of the diff for the same wording too, the other `legacy` mentions are transaction versions and the renamed `gmsol-legacy` binary, so those stay.

